PR #17, opened 2024-11-08
Save statement implementation

ascandone:
TODOs:
* [X] parse "save" syntax
* [X] handle statement in interpreter
* [x] handle statement in static analysis


ascandone:
Adding @altitude and @gfyrag as reviewers because, even though the PR is in still draft, the next part of the PR won't change the currently commited code

internal/analysis/hover_test.go
```diff
@@ -35,6 +35,54 @@ send $amt (
...
+	input := `vars { monetary $amt }
+
+save $amt from @acc
+`
+
+	rng := parser.RangeOfIndexed(input, "$amt", 1)
+
+	program := parser.Parse(input).Value
+	hover := analysis.HoverOn(program, rng.Start)
+	assert.NotNil(t, hover)
```
gfyrag, line 48: You should use require.NotNil instead of assert.NotNil.
If the assumtion fail, the code will continue, and trigger a panic later.

Same almost everywhere.

gfyrag (approved):

PR #26, opened 2024-11-26
feat: add featureFlags param

ascandone:
this is noop PR, but it involves changing the publicly exposed function, which we'll want to design in a way that will be backwards compatible in the future

### DIFF
--- a/internal/interpreter/interpreter.go
+++ b/internal/interpreter/interpreter.go
@@ -175,6 +175,7 @@ func RunProgram(
 	program parser.Program,
 	vars map[string]string,
 	store Store,
+	featureFlags map[string]struct{},
 ) (*ExecutionResult, InterpreterError) {
 	st := programState{
 		ParsedVars:         make(map[string]Value),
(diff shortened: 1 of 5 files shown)


numscript.go
```diff
@@ -56,7 +56,20 @@ type (
 )
 
 func (p ParseResult) Run(ctx context.Context, vars VariablesMap, store Store) (ExecutionResult, InterpreterError) {
-	res, err := interpreter.RunProgram(ctx, p.parseResult.Value, vars, store)
+	return p.RunWithFeatureFlags(ctx, vars, store, nil)
+}
+
+func (p ParseResult) RunWithFeatureFlags(
```
ascandone, line 62: Here we add a new publicly exposed function which has a `featureFlags map[string]struct{}` param
We have a map with untyped keys instead of a struct so that removing feature flags is not a breaking change

internal/interpreter/interpreter.go
```diff
@@ -175,6 +175,7 @@ func RunProgram(
 	program parser.Program,
 	vars map[string]string,
 	store Store,
+	featureFlags map[string]struct{},
```
ascandone, line 178: we pass the feature flags here directly (without further design patterns e.g. dependency injection). When a fflag will be available, we will parse it as a boolean field in the `programState` struct

this way
1. we can still control presence/absence of fflags in tests
2. the runtime will have access to passed feature flags in each moment

gfyrag (approved):

PR #28, opened 2024-12-06
Implemented string_to_account() function expression

ascandone:
This function implements two distinct functionalities:
1. overloading the "+" operator with strings concatenation
   1. e.g. allowing "ab" + "cd"  (=> "abcd")
2. create a `string_to_account` function which can be used to cast strings into account

## Feature flags
feature 2) is under the `--experimental-string-to-account-function` feature flag.
The string concatenation feature **IS NOT** under a feature flag. This means merging this PR will be equivalent as considering this functionality stable (in the numscript rewrite). This is discussed in the next paragraph

## Design consideration for string concatenation
It's safe to assume current design of string concatenation is good enough to be merged, e.g. we won't need any breaking changes.
While operator overloading's cost is to prevent a fully fledged type inference,
1. we already have operator overload for monetary+monetary and number+number operations
2. type inference is not needed as long as we have mandatory type annotations (e.g. in the "vars" block), and it's a good idea to maintain them mandatory for the sake of explicitness, and in order to allow the runtime not to depend on the static analysis (so that there isn't any perf overhead)
Other further development of the functionality will need more considerations, e.g. either forbidding `"ab" + 2` (current behaviour) or casting `2` to a string. The latter is, however, not a breaking change

## Open ended questions
1. should the function make sure numscript's grammar for account names is respected?


ascandone:
closing in favour of the account interpolation syntax

PR #29, opened 2024-12-20
Implemented underscore syntax in numbers

ascandone:


internal/parser/parser_test.go
```diff
@@ -43,6 +43,14 @@ func TestVariable(t *testing.T) {
 	snaps.MatchSnapshot(t, p.Value)
 }
 
+func TestNumberSyntaxUnderscore(t *testing.T) {
+	p := parser.Parse(`
+set_tx_meta("k1", 1_2_34_567)
```
ascandone, line 48: Here's the syntax

internal/parser/__snapshots__/parser_test.snap
```diff
@@ -2264,3 +2264,40 @@ parser.Program{
...
+                        End:   parser.Position{Character:16, Line:1},
+                    },
+                    String: "k1",
+                },
+                &parser.NumberLiteral{
+                    Range: parser.Range{
+                        Start: parser.Position{Character:18, Line:1},
+                        End:   parser.Position{Character:28, Line:1},
+                    },
+                    Number: 1234567,
```
ascandone, line 2297: Here's the resulting syntax tree

gfyrag (approved):

PR #35, opened 2025-02-04
change asset token definition

ascandone:
this PR narrows the available asset values, forbidding assets like "/", "EUR/", or  "/42"
note this is a breaking change (but the old behaviour was not intented and therefore not documented)


Numscript.g4
```diff
@@ -40,7 +40,7 @@ IDENTIFIER: [a-z]+ [a-z_]*;
 NUMBER: MINUS? [0-9]+ ('_' [0-9]+)*;
 VARIABLE_NAME: '$' [a-z_]+ [a-z0-9_]*;
 ACCOUNT: '@' [a-zA-Z0-9_-]+ (':' [a-zA-Z0-9_-]+)*;
-ASSET: [A-Z/0-9]+;
+ASSET: [A-Z]+ ('/' [0-9]+)?;
```
laouji, line 43: Is `[A-Z]+` specific enough or should we maybe match exactly 3 with `{3}`. Not sure if currency is ever more or less than 3 (maybe user defined ones?)

ascandone: you mean the asset's length? Yeah It may be an arbitrary asset like GEM, COIN, etc
Potentially also something  like some non fungible asset like `TKEA2BC34EE0FC4E3DAD46874762279B13` - which I just realized we cannot express anymore after this commit. I already asked @altitude  if this grammar change was correct but I'll double check explicitly this

ascandone: UPDATE: looking at the ledger's checks, I've relaxed this constraint so that assets can have numbers too (just not start with a number)
This is still more relaxed than ledger's (variables) checks, which allow a max of 16 chars for asset and 6 chars for the "/" part. 
ledger's regex: [A-Z][A-Z0-9]{0,16}(\/\d{1,6})?
@gfyrag @altitude should we lower down the characters numbers (this is a breaking change for numscript) or relax ledger's regex?

gfyrag: The format used by the ledger and by the original machine is `[A-Z][A-Z0-9]{0,16}(/\d{1,6})?`
I guess Numscript should be aligned on this format.

gfyrag: I realize the grammar allow `ASSET: [A-Z/0-9]+;`
But the code check the format at execution and check for the format `[A-Z][A-Z0-9]{0,16}(/\d{1,6})?`.

PR #38, opened 2025-02-16
perf: make lsp handling concurrent

ascandone:
This PR wraps a bunch of lines in a goroutine, so that while an expensive LSP request is processed, other requests can be taken care of in the meanwhile
A mutex is used in order to make sure that we don't write to stdout concurrently (producing invalid messages)

I don't have tooling to measure performance improvement, and this optimization is probably not going to matter for now, but it may be useful as we iterate with more complex LS features


internal/lsp/server.go
```diff
@@ -37,24 +38,31 @@ func SendNotification(method string, params interface{}) {
 
 func RunServer[State any](args ServerArgs[State]) {
 	buf := NewMessageBuffer(os.Stdin)
+	mu := sync.Mutex{}
 
 	for {
 		request := buf.Read()
```
gfyrag, line 44: Are you sure the request object is thread sage?

internal/lsp/server.go
```diff
@@ -37,24 +38,31 @@ func SendNotification(method string, params interface{}) {
...
+				panic(err)
+			}
+
+			rawMsg := json.RawMessage(bytes)
+			encoded := encodeMessage(jsonrpc2.Response{
+				ID:     request.ID,
+				Result: &rawMsg,
+			})
+
+			mu.Lock()
```
gfyrag, line 58: Why take a lock before doing a fmt.Print?

gfyrag (commented):
I suggest you to read code rabbit insight about error handling.

internal/lsp/server.go
ascandone: In order to avoid concurrent writes to stdout which would mix the outputs, thus making the response(s) unreadable
(Context: language server protocol communicates via a json-rpc API over stdio and stdout)

internal/lsp/server.go
ascandone: I cannot see why not
The `request` object is just a plain object which is only read (is the result of the parsing of a json-rpc input payload)
besides, it's accessed by only one goroutine anyway

gfyrag: You have :
```
bytes, err := json.Marshal(args.Handler(request, &args.InitialState))
if err != nil {
	panic(err)
}
```
in the go routine, so the object is accessible by multiple go routine.
But if the object is only read, I guess it is ok.

internal/lsp/language_server/server.go
```diff
@@ -0,0 +1,71 @@
...
+	writeMutex.Lock()
+	os.Stdout.Write([]byte(encoded))
+	writeMutex.Unlock()
+}
+
+func RunServer[State any](args ServerArgs[State]) {
+	buf := newMessageBuffer(os.Stdin)
+	mu := sync.Mutex{}
+
+	state := args.InitialState(func(method string, params any) {
```
ascandone, line 38: This file was copied-paste and moved in its own module.
The main difference is that the "sendNotification" function, instead of being called directly, is stored by the server consumer (the lsp module) in its state, so that the handlers can call that instead.
The instance of the state will be passed as value, not reference, in order to discourage writing there mutable state directly, which will be accessed concurrently.
The consumer will have to handle concurrent access manually by storing pointers explicitly

internal/lsp/documents_store.go
```diff
@@ -0,0 +1,28 @@
+package lsp
+
+import "sync"
+
+type documentStore[Doc any] struct {
```
ascandone, line 5: I'm aware that a `sync.Map` data structure exists, but this one looks more "fine grained", and its simple enough

gfyrag (approved):

PR #45, opened 2025-03-05
Feat/expr in var origin

ascandone:
NOTE FOR REVIEWERS: commits are rubbish. The PR is not meant to be read commit by commit, and it'll be squashed before merge (as always, discard the antlr autogenerated part)

## Specs
This PR changes the numscript grammar so that the variable origin is not just a function call but any expression. Conversely, a function call is now an expression.
This means that:
```
vars {
  number $b = 1 + 2
}
```
is now possible, and at the same time:
```
send balance(@alice, 42) { .. }
```
is now allowed as well

**important caveat:** it will only be possible to call the meta() function in a toplevel position of a var origin (same way as it was until this PR). This is because otherwise it would be tricky (and even trickier for the user) to determine the return type (which is needed to parse correctly the stringified meta)

**feature flags:** the feature is under the `experimental-mid-script-function-call` feature flag


internal/interpreter/balances.go
```diff
@@ -0,0 +1,57 @@
...
+	for accountName, queriedCurrencies := range q {
+		for _, currency := range queriedCurrencies {
+			isAlreadyCached := b.has(accountName, currency)
+			if !isAlreadyCached {
+				filteredQuery[accountName] = queriedCurrencies
+			}
+		}
+	}
+	return filteredQuery
+}
```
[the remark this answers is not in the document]
ascandone: damn it, this time you were right

internal/parser/ast.go
```diff
@@ -18,6 +18,7 @@
 func (*NumberLiteral) valueExpr()        {}
 func (*StringLiteral) valueExpr()        {}
 func (*BinaryInfix) valueExpr()          {}
+func (*FnCall) valueExpr()               {}
```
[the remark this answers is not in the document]
ascandone: that's an empty method, only used to define what's implementing the interface

internal/interpreter/interpreter_error.go
```diff
@@ -220,3 +220,11 @@ type InvalidAccountName struct {
 func (e InvalidAccountName) Error() string {
 	return fmt.Sprintf("Invalid account name: @%s", e.Name)
 }
+
+type InvalidNestedMeta struct {
```
gfyrag, line 224: You should implements `Is(target error) bool` method for your errors.
It allow callers (the ledger for example) to check for concrete errors.

gfyrag: See https://pkg.go.dev/errors

internal/interpreter/interpreter_test.go
```diff
@@ -2218,6 +2252,67 @@ func TestVariableBalance(t *testing.T) {
 // 	test(t, tc)
 // }
 
+func TestBalanceSimple(t *testing.T) {
+	script := `
```
gfyrag, line 2256: add t.Parallel() call on your tests except if it's not parallelizable, but I guess it is the case here

internal/utils/utils.go
```diff
@@ -32,3 +32,13 @@ func MaxBigInt(a *big.Int, b *big.Int) *big.Int {
 
 	return &max
 }
+
+func Filter[T any](slice []T, predicate func(x T) bool) []T {
```
gfyrag, line 36: You have some helpers in go libs to play with generics.
https://github.com/formancehq/go-libs/tree/main/collectionutils

gfyrag (approved):

internal/utils/utils.go
ascandone: yeah I'm aware of this lib but for such simple utilities that are basically one-liners it's just better to duplicate them
(I've also tried to avoid this so far because I though it's sort of non-idiomatic in go, but it's just so much clearer this way..)

internal/interpreter/interpreter_test.go
ascandone: oh yeah that's true.. yeah I think so far 100% of the tests are parallelizable, but I keep forgetting about adding it.. is there a way to declare that the whole module is parallelizable ?

PR #48, opened 2025-03-07
feat: add option to cli

ascandone:


altitude (approved):

Varun786223:
@ascandone

Hey there, loving this CLI tweak—adding a flag for account interpolation is a neat touch! I’ve given it a once-over with my battle-scarred dev eyes, and it’s looking solid. No crashes lurking, but a few spots could use a quick shine before we hit merge. Here’s my take:

Global Flag Scope
accountInterpolationFlag as a global gets the job done, but globals give me the heebie-jeebies from past messes. Could we pass it from getRunCmd to run instead? Keeps it clean if the CLI sprouts more features. Not a dealbreaker, just a “future-proof” nudge.
PR Description
Blank slate here—what’s account interpolation all about? Why’s it experimental? A quick blurb (e.g., “Toggles fancy account string parsing, still in testing”) would set the stage and sell your win.
Coverage Gap
Codecov’s grumbling about 0% on those 4 lines—fair call for new code. A simple test (e.g., flag on, check interpreter.ExperimentalAccountInterpolationFlag) would shut it up and catch any gremlins. Five-minute job, max.
Flag Naming
Hoping it’s something obvious like --account-interpolation. If it’s vaguer (e.g., --exp), maybe clarify it for us mortals? Just a sanity check.

PR #52, opened 2025-03-24
LS rewrite

ascandone:
_NOTE FOR REVIEWERS_: this PR is not designed to  be read commit by commit.
You can follow modules dependency graph instead:
1. internal/jsonrpc2/*
2. internal/lsp/*

---

This PR doesn't add any new functionality for the user but repays a bit of the tech debt of the previous lsp implementation. The main issue is that previous implementation had 0 test coverage, except for the data decoding/encoding part

* rewrite LSP from scratch
* removes sourcegraph/jsonrpc2 dependency, which was only used to (de)serialize rpc messages, thus removing transitive dependencies like gorilla/websocket
* Improve ability to test the json-rpc and language server implementations
* separate clearly LS-specific, json-rpc-specific, and data encoding parts. This also makes the json-rpc implementation more flexible
* most importantly, the implementation is now more solid and complete. E.g. the server can now receive notifications ([example](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration)) and send requests ([example](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration)).
  * as a side effect, the ls json-rpc implementation can now be used symmetrically as client as testing utility

I kept a from-scratch implementation of json-rpc2 because 1. other alternatives are still marked as unstable (examples: [sourcegraph/jsonrpc2](https://pkg.go.dev/github.com/sourcegraph/jsonrpc2@v0.2.0#section-readme), [go/exp/jsonrpc2](https://pkg.go.dev/golang.org/x/exp/jsonrpc2)) 2. the other implementation look more low-level and complex
so it though made sense not to depend on those for a 200 LOC implementation

This PR _doesn't_ handle some jsonrpc2 features which we aren't used from language servers, like messages batching, as well as specs-compliant error handling of every corner case, as we'll rely for the client's to honor the protocol

### DIFF
--- a/internal/lsp/handlers.go
+++ b/internal/lsp/handlers.go
@@ -259,6 +259,9 @@ func NewConn(objStream jsonrpc2.MessageStream) *jsonrpc2.Conn {
 			state.updateDocument(conn, p.TextDocument.URI, p.TextDocument.Text)
 		}),
 		jsonrpc2.NewNotificationHandler("textDocument/didChange", jsonrpc2.SyncHandling, func(p lsp_types.DidChangeTextDocumentParams, conn *jsonrpc2.Conn) {
+			if len(p.ContentChanges) == 0 {
+				return
+			}
 			text := p.ContentChanges[len(p.ContentChanges)-1].Text
 			state.updateDocument(conn, p.TextDocument.URI, text)
 		}),


internal/analysis/document_symbols.go
```diff
@@ -31,8 +33,15 @@ func (r *CheckResult) GetSymbols() []DocumentSymbol {
 			Range:          v.Name.Range,
 			SelectionRange: v.Name.Range,
 		})
-
 	}
 
+	slices.SortFunc(symbols, func(a, b DocumentSymbol) int {
+		if a.Range.Start.GtEq(b.Range.Start) {
```
gfyrag, line 39: not stable sort ?

ascandone: the order doesn't matter match. In fact, they weren't even sorted initially - it's just more handy to have a deterministic way to be outputted for testing purposes (they are taken from a map, whose order is not reliable)

ascandone: (and for this reason, stable sort wouldn't be "better" than regular sorting)

internal/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,279 @@
...
+							ID:    id,
+							Error: &ErrInvalidParams,
+						})
+						return
+					}
+				}
+
+				out := handler(payload, conn)
+
+				bytes, _ := json.Marshal(out)
```
gfyrag, line 66: you should not ignore it.
imho, panic, if you don't know how to handle it.

internal/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,279 @@
...
+			err = conn.handleMessage(msg)
+			if err != nil {
+				conn.listenErr <- err
+				return
+			}
+		}
+	}()
+
+	return &conn
+}
```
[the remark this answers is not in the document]
ascandone: I'm not sure about this: .Close() is not meant to be called many times concurrently

internal/jsonrpc2/chan_object_stream.go
```diff
@@ -0,0 +1,31 @@
+package jsonrpc2
+
+// This is mostly intented as a testing utility
```
ascandone, line 3: this is only used in tests so it doesn't matter if the implementation is inaccurate

internal/lsp/handlers.go
```diff
@@ -135,24 +127,24 @@ func (state *State) handleGotoDefinition(params DefinitionParams) *Location {
...
-func (state *State) handleGetSymbols(params DocumentSymbolParams) []DocumentSymbol {
+func (state *State) handleGetSymbols(params lsp_types.DocumentSymbolParams) []lsp_types.DocumentSymbol {
 	doc, ok := state.documents.Get(params.TextDocument.URI)
 	if !ok {
 		return nil
 	}
 
-	var lspDocumentSymbols []DocumentSymbol
+	var lspDocumentSymbols []lsp_types.DocumentSymbol
 	for _, sym := range doc.CheckResult.GetSymbols() {
```
gfyrag, line 143: Do you expect a lot of symbols?

internal/jsonrpc2/jsonrpc2_test.go
```diff
@@ -0,0 +1,76 @@
...
+		jsonrpc2.NewRequestHandler("sum", jsonrpc2.AsyncHandling, func(p SumParams, conn *jsonrpc2.Conn) any {
+			return p.X + p.Y
+		}),
+	)
+
+	raw, err := client.SendRequest("sum", SumParams{X: 100, Y: 42})
+	require.Nil(t, err)
+
+	var res int
+	json.Unmarshal(raw, &res)
```
gfyrag, line 27: check error

internal/jsonrpc2/jsonrpc2.go
```diff
@@ -0,0 +1,282 @@
...
+		},
+	}
+}
+
+// Starts listening asynchronously to the MessageStream and returns the connection.
+//
+// By default, the server will try write concurrently to the ObjectStream
+func NewConn(objStream MessageStream, handlers ...Handler) *Conn {
+	conn := Conn{
+		listenErr:            make(chan error),
```
gfyrag, line 133: If no listener are listening this channel, your loop will block.
You can create a chan of size 1 to avoid that.
```
make(chan error, 1)
```

internal/lsp/object_stream.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if err2 != nil {
+		return err2
+	}
+
+	return nil
+}
+
+func (s *LsObjectStream) WriteMessage(obj jsonrpc2.Message) error {
+	bytes, err := json.Marshal(obj)
```
gfyrag, line 50: not a big deal but you can avoid some memory consumtion by streaming the marshaling.
```
json.NewEncoder(s.out).Encode(obj)
```

gfyrag (approved):

PR #53, opened 2025-04-03
breaking: pass flags as slice instead of many args

ascandone:
As we're creating more and more experimental features, it's becoming annoying to have a cli arg for each flag (Don't worry, we'll start actually making them stable soon - probably as numscript interpreter itself becomes stable)
So we'll pass them as a slice instead

Note: even though we created a way to "pass" flags via comment annotation  with this PR https://github.com/formancehq/numscript/pull/51 , the PR's idea is that those comments are meant for static analysis (that is, "numscript check" and the vscode extension). During the execution phase, those comments are treated as comments - that is, ignored entirely
I get that this might be considered a bit confusing and inconsistent, so this decision is opened for feedback. We can just use the `// @ feature_flag my-flag` comments as the sole feature flag mechanism instead
One reason why that's not done yet, is that if we were to rely on those for execution too, we'd have to define more precisely the behaviour; for example:
* defining the grammar more precisely (e.g. is `// @ feature_flag my-flag and other trailing random text` a valid annotation? )
* where are those comments legal? can they be at the end of the file? Can they be not at the start of the line?
Those issues are less serious if the annotations are just for vscode diagnostics, but if users were to use them to run actual code, such a flaky mechanism wouldn't be an idea API: it'd probably be better to pass flags explicitly to the cli, or the ledger config, or the function call.

To recap:
* static analysis => flags as comments
* interpreter runtime => flags passed explicitly (comments are ignored)

### DIFF
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -21,14 +21,15 @@ const (
 	OutputFormatJson   = "json"
 )
 
-var runVariablesOpt string
-var runBalancesOpt string
-var runMetaOpt string
-var runRawOpt string
-var runStdinFlag bool
-var runOutFormatOpt string
-
-var flags_ []string
+type Args struct {
+	VariablesOpt string
+	BalancesOpt  string
+	MetaOpt      string
+	RawOpt       string
+	StdinFlag    bool
+	OutFormatOpt string
+	Flags        []string
+}
 
 type inputOpts struct {
 	Script    string                       `json:"script"`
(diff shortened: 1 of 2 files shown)


internal/cmd/run.go
```diff
@@ -213,11 +202,9 @@ func getRunCmd() *cobra.Command {
 	cmd.Flags().BoolVar(&runStdinFlag, "stdin", false, "Take input from stdin (same format as the --raw option)")
 
 	// Feature flag
-
-	cmd.Flags().BoolVar(&overdraftFeatureFlag, flags.ExperimentalOverdraftFunctionFeatureFlag, false, "enables the experimental overdraft() function")
-	cmd.Flags().BoolVar(&oneOfFeatureFlag, flags.ExperimentalOneofFeatureFlag, false, "enable the experimental oneof combinator")
-	cmd.Flags().BoolVar(&accountInterpolationFlag, flags.ExperimentalAccountInterpolationFlag, false, "enables an account interpolation syntax, e.g. @users:$id:pending")
-	cmd.Flags().BoolVar(&midScriptFunctionCallFeatureFlag, flags.ExperimentalMidScriptFunctionCall, false, "allows to use function call as expression, and to use any expression when definining variables")
+	cmd.Flags().StringSliceVar(&flags_, "flags", nil, fmt.Sprintf("the feature flags to pass to the interpreter. Currently available flags: %s",
```
gfyrag, line 205: Very bad global vars.
If you try to test while parallelizing, it will probably break.

gfyrag (approved):

PR #76, opened 2025-07-14
Test/migrate tests with script

ascandone:


internal/interpreter/interpreter_test.go
```diff
@@ -1,21 +1,37 @@
...
 
 	"testing"
 
 	"github.com/formancehq/numscript/internal/parser"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+const scriptsFolder = "../../testdata/script-tests"
```
Dav-14, line 20: - can you move the testdata folder inside the internal/interpreter to avoid relative path ?

internal/interpreter/interpreter_test.go
```diff
@@ -1,21 +1,37 @@
...
+
+func TestScripts(t *testing.T) {
```
Dav-14, line 22: Is there other test ? you might want to use t.Parrallel()

Does `RunSpecs` use parralelism ?

ascandone:
sorry, I missed the comments and merged it anyway. I'll address them in the next PR 👀

PR #77, opened 2025-07-15
feat: Use Nix & Shipfox & Just for DevXP

flemzord:


.github/workflows/main.yml
```diff
@@ -0,0 +1,112 @@
...
+        with:
+          fetch-depth: 0
+      - name: Setup Env
+        uses: ./.github/actions/default
+        with:
+          token: ${{ secrets.NUMARY_GITHUB_TOKEN }}
+      - run: >
+          nix develop --impure --command just tests
+        env:
+          SPEAKEASY_API_KEY: ${{ secrets.SPEAKEASY_API_KEY }}
```
ascandone, line 71: do we need that?

PR #97, opened 2025-10-07
feat: MCP first MCP

ascandone:


internal/mcp_impl/handlers.go
```diff
@@ -0,0 +1,195 @@
...
+			return mcpErr, nil
+		}
+
+		vars, mcpErr := parseVarsJson(request.GetArguments()["vars"])
+		if mcpErr != nil {
+			return mcpErr, nil
+		}
+
+		out, iErr := interpreter.RunProgram(
+			context.Background(),
```
Dav-14, line 124: You might want to use the proper context or a derived one ?

ascandone: This one should be ok: we are only doing in-memory operations, and fetching data in a go struct
(We need context because the implementor of the interface might be sql queries )

gfyrag: David is right. 
Always use the correct context. 
There is no good reason to not use it.

PR #102, opened 2025-11-13
feat: asset scaling

ascandone:


internal/interpreter/interpreter.go
```diff
@@ -617,6 +679,59 @@ func (s *programState) trySendingUpTo(source parser.Source, amount *big.Int) (*b
...
+	case *parser.SourceWithScaling:
+		account, err := evaluateExprAs(s, source.Address, expectAccount)
+		if err != nil {
+			return nil, err
+		}
+
+		baseAsset, assetScale := getAssetScale(s.CurrentAsset)
+		acc, ok := s.CachedBalances[*account]
+		if !ok {
+			panic("TODO accountBal not found")
```
ascandone: Those TODOs happen when there's a @world account. I'm not sure what we should do in this case; I'm considering either preventing that (statically and on runtime), or treat it as if "with scaling" was a noop

Numscript.g4
```diff
@@ -43,10 +43,10 @@ allotment:
 
 colorConstraint: RESTRICT valueExpr;
 
-source:
-	address = valueExpr colorConstraint? ALLOWING UNBOUNDED OVERDRAFT						# srcAccountUnboundedOverdraft
-	| address = valueExpr colorConstraint? ALLOWING OVERDRAFT UP TO maxOvedraft = valueExpr	#
-		srcAccountBoundedOverdraft
+source
+	: address = valueExpr colorConstraint? ALLOWING UNBOUNDED OVERDRAFT						# srcAccountUnboundedOverdraft
+	| address = valueExpr colorConstraint? ALLOWING OVERDRAFT UP TO maxOvedraft = valueExpr	#srcAccountBoundedOverdraft
```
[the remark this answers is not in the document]
ascandone: we'll ignore it for now as it's not about this PR

internal/interpreter/asset_scaling.go
```diff
@@ -0,0 +1,135 @@
...
+		if intPart.Cmp(big.NewInt(0)) == 0 {
+			continue
+		}
+
+		out = append(out, scalePair{
+			scale:  p.scale,
+			amount: intPart,
+		})
+
+		if leftAmt != nil && leftAmt.Cmp(big.NewInt(0)) != 1 {
```
Azorlogh: nit: can use .Sign() here

internal/interpreter/asset_scaling.go
```diff
@@ -0,0 +1,135 @@
...
+			break
+		}
+	}
+
+	return out, totalSent
+}
```
Azorlogh: This is returning an empty []scalePair, afaiu it should return `[{scale: 0, amount: 1}]`
```
sol, tot := findSolution(
		big.NewInt(2),
		2,
		map[int64]*big.Int{
			0: big.NewInt(1),
		})
```

ascandone:
@corerabbitai review

Azorlogh (approved):
Went through everything, new behavior looks good to me!
Great work 💪

PR #106, opened 2025-12-08
fix: save-all when balance is negative

Azorlogh:
## Summary by cubic
Fixes save-all so negative balances are not reset to zero. This prevents incorrect fund availability and correctly triggers missingFunds for overdrawn accounts.

- **Bug Fixes**
  - Only zero out balance on save-all when the balance is positive.
  - Added a script test for a negative USD balance that expects missingFunds.

<sup>Written for commit 23909b208a7f253e174b95404468b07198768562. Summary will update automatically on new commits.</sup>


ascandone (approved):
thanks!

PR #109, opened 2026-01-31
fix: fix float perc syntax bugs

ascandone:
There were 2 bugs in the syntax of % with floating points:
1. `0.19%` wasn't using base 10. We switch "base=0" with "base=10" to make sure this always happens
2. `0.01%` wasn't taking into account the leading zero and was being considered the same as `1%`


internal/parser/parser.go
```diff
@@ -255,23 +255,41 @@ func parseSource(sourceCtx antlrParser.ISourceContext) Source {
...
 		return nil, 0, err
 	}
 
 	var floatingDigits uint16
 	split := strings.Split(str, ".")
+
 	if len(split) > 1 {
-		floatingDigits = uint16(len(split[1]))
+		floatingPart := split[1]
+		floatingDigits = uint16(len(floatingPart)) + countLeadingZeros(floatingPart)
```
Azorlogh: I think this is wrong - floatingPart is already a string so already contains the floating zeros.

So this counts the leading zeros twice. You can see it with this test:
sending `0.019%` of `[EUR 100_000]`

Expected: 19, got: 2

PR #116, opened 2026-02-17
refactor: change funds queue internals

ascandone:
Change the funds queue internals by using a slice instead of a stack implementation
Credits: @Azorlogh (https://github.com/formancehq/numscript/pull/100)


flemzord:
Closed automatically: bulk stale PR cleanup

PR #129, opened 2026-03-27
fix: limit decimal precision in ParsePercentageRatio to 18 places

flemzord:
## Summary
- Fixed a bug discovered by fuzz testing where `ParsePercentageRatio` accepted unreasonably large decimal precision (100+ digits)
- Added validation to limit decimal places to 18, which is sufficient for financial calculations and matches common cryptocurrency precision standards
- Added a fuzz test to prevent regression

## Bug Details
The fuzzer discovered that inputs like `"0.00000...000%"` with 100+ decimal digits were accepted and returned `floatingDigits > 100`. This could potentially cause:
- Integer overflow when casting to uint16
- Performance issues with extremely large scale values
- Unexpected behavior in downstream code expecting reasonable precision

## Fix
Added a check in `ParsePercentageRatio` to reject percentages with more than 18 decimal places with a clear error message.

## Test plan
- [x] Fuzz test runs for 120 seconds without finding issues
- [x] All existing tests pass
- [x] Linter passes
- [x] Added FuzzParsePercentageRatio to prevent regression


ascandone (changes requested):
I'd rather going with this approach instead
https://github.com/formancehq/numscript/pull/130

PR #135, opened 2026-05-18
feat: add ResolveDependencies for concrete volume/metadata discovery

gfyrag:
## Context

Ledger v3 uses a Raft architecture where transactions go through two stages:
1. **Admission** (leader): discovers script dependencies, preloads volumes, proposes a Raft order
2. **FSM** (all nodes): applies the order deterministically

The admission layer needs to know which volumes to preload **before** executing the script — but the volumes depend on the script itself.

The current approach uses `GetInvolvedAccounts` which returns expression trees (`InvolvedAccountExpr`) that the consumer must resolve. This led to ~500 lines of expression tree walking code on the ledger side, with bugs (GetAmount/GetAsset extracting wrong component, IsValidCall never called) and limitations (balance-dependent account addresses rejected).

## Proposal

Add `ResolveDependencies()` which returns **concrete data** instead of expression trees:

```go
type ResolvedDependencies struct {
    Volumes  map[string]map[string]*big.Int  // account → asset → balance
    Metadata map[string]map[string]string    // account → key → value
}

func (p ParseResult) ResolveDependencies(ctx, vars, store) (*ResolvedDependencies, error)
```

The consumer provides a `Store` callback (same interface as `Run`), gets back concrete reads, and handles the rest (hashing, preloading, drift detection) on its side.

## How the ledger uses it

**Admission:** call `ResolveDependencies` → get volumes + metadata → hash inputs → preload into Raft order

**FSM:** call `ResolveDependencies` again → hash inputs → compare with admission hash → if match, execute; if mismatch, reject (retryable)

This is classic optimistic concurrency control.

## What this replaces on the ledger side

- ~500 lines of `resolveExprToString`, `resolveExprWithMetadata`, `collectFnMetaReads`, `ResolveDeferredAccounts`
- No more coupling with `InvolvedAccountExpr` types
- No more `IsValidCall` / non-deterministic script rejection
- All features (meta chains, balance in addresses, mid-script calls) work transparently

## Implementation

- `recordingStore` wraps a `Store` and records all `GetBalances`/`GetAccountsMetadata` calls
- `ResolveDependencies` runs the script with the recording store, discards the execution result, returns the recorded reads
- The internal implementation can be optimized later (partial execution, static analysis) without changing the API

## Test plan

- [x] Simple transfer — records source balance
- [x] World source — no balance recorded
- [x] meta() call — records metadata read
- [x] Multiple sources — records all source balances
- [x] Variables — records resolved account balances
- [x] balance() function — records balance read
- [x] Multiple sends — records all reads
- [x] set_account_meta — no metadata reads produced
- [x] Nested meta() chain — records all chained metadata reads
- [x] Send all (*) — records source balance
- [x] Empty reads — world-only script produces no reads
- [x] Full existing test suite passes

### DIFF
--- a/internal/interpreter/resolve_dependencies.go
+++ b/internal/interpreter/resolve_dependencies.go
@@ -1,0 +1,114 @@
+package interpreter
+
+import (
+	"context"
+	"maps"
+	"math/big"
+	"slices"
+
+	"github.com/formancehq/numscript/internal/flags"
+	"github.com/formancehq/numscript/internal/parser"
+)
+
+// ResolvedDependencies holds the concrete data a script reads during resolution.
+// The consumer can use this to preload volumes and detect input drift.
+type ResolvedDependencies struct {
+	// Volumes contains all (account, asset) → balance pairs read during resolution.
+	Volumes map[string]map[string]*big.Int
+
+	// Metadata contains all (account, key) → value pairs read during resolution.
+	Metadata map[string]map[string]string
+}
+
+// ResolveDependenciesOptions configures ResolveDependencies behavior.
+type ResolveDependenciesOptions struct {
+	// FeatureFlags enables additional experimental features (same as RunWithFeatureFlags).
+	FeatureFlags map[string]struct{}
+
+	// ForbiddenFlags rejects scripts that declare any of these features.
+	// This takes precedence over script-level #![feature("...")] directives.
+	// Use this to block features that ResolveDependencies cannot fully resolve
+	// (e.g. experimental-mid-script-function-call).
+	ForbiddenFlags map[string]struct{}
+}
+
+// ResolveDependencies discovers which balances and metadata a script will read
+// by performing variable resolution and balance preloading — the same two phases
+// that RunProgram does before executing statements. It does NOT execute the
+// statements themselves (no postings are produced).
+//
+// This covers all store reads for scripts that don't use
+// experimental-mid-script-function-call. Scripts using that feature may trigger
+// additional balance reads during execution (e.g. balance() called between two
+// send statements, where the result depends on the first send's postings).
+// Consumers that cannot tolerate incomplete dependency lists should forbid this
+// feature via ForbiddenFlags.
+func ResolveDependencies(
+	ctx context.Context,
+	program parser.Program,
+	vars map[string]string,
+	store Store,
+	opts ResolveDependenciesOptions,
+) (*ResolvedDependencies, InterpreterError) {
+	recorder := newRecordingStore(store)
+
+	featureFlags := maps.Clone(opts.FeatureFlags)
+	if featureFlags == nil {
+		featureFlags = make(map[string]struct{}, len(program.Flags))
+	}
+
+	for _, flag := range program.Flags {
+		index := slices.Index(flags.AllFlags, flag.String)
+		if index == -1 {
+			return nil, InvalidFeature{Feature: flag.String}
+		}
+
+		if _, forbidden := opts.ForbiddenFlags[flag.String]; forbidden {
+			return nil, ForbiddenFeature{FlagName: flag.String}
+		}
+
+		featureFlags[flag.String] = struct{}{}
+	}
+
+	// Replicate the initialization and preload phases of RunProgram,
+	// but stop before statement execution.
+	st := programState{
+		ParsedVars:         make(map[string]Value),
+		TxMeta:             make(map[string]Value),
+		CachedAccountsMeta: AccountsMetadata{},
+		CachedBalances:     Balances{},
+		SetAccountsMeta:    AccountsMetadata{},
+		Store:              recorder,
+		Postings:           make([]Posting, 0),
+		fundsQueue:         newFundsQueue(nil),
+
+		CurrentBalanceQuery: BalanceQuery{},
+		ctx:                 ctx,
+		FeatureFlags:        featureFlags,
+	}
+
+	// Phase 1: parse variables — resolves meta(), balance(), overdraft() origins.
+	st.varOriginPosition = true
+	if program.Vars != nil {
+		if err := st.parseVars(program.Vars.Declarations, vars); err != nil {
+			return nil, err
+		}
+	}
+	st.varOriginPosition = false
+
+	// Phase 2: traverse statement ASTs to discover balance needs, then preload.
+	for _, statement := range program.Statements {
+		if err := st.findBalancesQueriesInStatement(statement); err != nil {
+			return nil, err
+		}
+	}
+
+	if err := st.runBalancesQuery(); err != nil {
+		return nil, QueryBalanceError{WrappedError: err}
+	}
+
+	return &ResolvedDependencies{
+		Volumes:  recorder.balanceReads,
+		Metadata: recorder.metadataReads,
+	}, nil
+}
--- a/internal/interpreter/resolve_dependencies.go
+++ b/internal/interpreter/resolve_dependencies.go
@@ -3,46 +3,63 @@ package interpreter
 import (
 	"context"
 	"maps"
-	"math/big"
 	"slices"
 
 	"github.com/formancehq/numscript/internal/flags"
 	"github.com/formancehq/numscript/internal/parser"
+	"github.com/formancehq/numscript/internal/utils"
 )
 
-// ResolvedDependencies holds the concrete data a script reads during resolution.
-// The consumer can use this to preload volumes and detect input drift.
+// ResolvedDependencies summarizes what a script reads from and writes to the
+// store. The caller can use it to preload data and to detect input drift
+// between successive runs.
 type ResolvedDependencies struct {
-	// Volumes contains all (account, asset) → balance pairs read during resolution.
-	Volumes map[string]map[string]*big.Int
+	// Reads contains the data the script read from the store while resolving.
+	Reads ResolvedReads
 
-	// Metadata contains all (account, key) → value pairs read during resolution.
-	Metadata map[string]map[string]string
+	// Writes contains the (account, asset, color) tuples whose balance can be
+	// impacted by a posting emitted by the script.
+	Writes ResolvedWrites
+}
+
+// ResolvedReads holds the data read from the store while resolving the
+// script's dependencies.
+type ResolvedReads struct {
+	// Volumes contains every (account, asset, color) → balance row read from
+	// the store, in the order it was returned.
+	Volumes Balances
+
+	// Metadata contains all (account, key) → value pairs read from the store.
+	Metadata AccountsMetadata
+}
+
+// ResolvedWrites holds the data the script may write to the store.
+type ResolvedWrites struct {
+	// Volumes lists every (account, asset, color) tuple that may be impacted
+	// by a posting emitted by the script.
+	Volumes BalanceQuery
 }
 
 // ResolveDependenciesOptions configures ResolveDependencies behavior.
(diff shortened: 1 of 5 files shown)


gfyrag:
## Note on `experimental-mid-script-function-call`

`ResolveDependencies` performs Phases 1-2 of `RunProgram` (variable resolution + balance preloading) but does **not** execute statements. This means it captures all store reads for standard scripts.

The exception is `experimental-mid-script-function-call`, which allows calling `balance()` **between** send statements:

```numscript
#![feature("experimental-mid-script-function-call")]

send [USD/2 100] (
  source = @world
  destination = @acc
)

// This balance() reads a value that depends on the first send's result.
// It cannot be discovered without actually executing the statements.
send balance(@acc, USD/2) (
  source = @acc
  destination = @dest
)
```

In this case, the mid-script `balance()` call produces a store read that depends on the execution of prior statements. `ResolveDependencies` cannot capture it because it doesn't execute statements.

To handle this, `ResolveDependenciesOptions` includes a `ForbiddenFlags` field. Consumers that need a **complete** dependency list can forbid this feature:

```go
deps, err := parsed.ResolveDependencies(ctx, vars, store, numscript.ResolveDependenciesOptions{
    ForbiddenFlags: map[string]struct{}{
        "experimental-mid-script-function-call": {},
    },
})
```

Scripts that declare this feature will be rejected with a `ForbiddenFeature` error, making the incompleteness explicit rather than silent.

internal/interpreter/resolve_dependencies_test.go
```diff
@@ -264,3 +264,64 @@ send balance(@acc, USD/2) (
...
+				"account_that_needs_balance": {
+					"USD/2": big.NewInt(42),
+				},
+			},
+			Meta: AccountsMetadata{"account_that_needs_meta": {"k": "source2"}},
+		})
+
+	require.Equal(t, deps.Volumes, map[string]map[string]*big.Int{
+		"source1": {
+			"COIN": big.NewInt(123),
```
gfyrag: Don't need the amounts, just need to know the touched volumes.
You can keep it if you're comfident with it, but it add you more surface to maintain.

ascandone: If we don't need the amount it's better to remove them so that the impl is more "conservative".

However the point here is that this test shows that we still output a "Volumes" result even if the set of involved (account,asset) pairs depends in the meta and/or balances themselves. (In the case of the test, just on the meta). We had talked about this a bit (my understanding is that, if it happens, you'll detect that and just retry the thing) but I'm asking again to double check if we are ok with this

gfyrag: Yes, it should work. I've not noted the missing metadata key in deps.
But it should work.
Is the resolution determinist? (same calls in the same order?)

gfyrag: Hum, sorry, wrong response.
The GetAccountMetadata will be exercised in the admission path AND in the hot path.
If the dependency resolution does not report the metadata, it will not be loaded in the raft command, and so, not loaded in the in memory storage of the hot path.

gfyrag: So, a totally valid transaction will be rejected because the metadata will not be found in ram.

gfyrag:
@ascandone updated after the rescope discussion — context for the new approach below.

## Rescope

Two concrete needs:
1. Make all store calls deterministic so the host can hash them and detect input drift between successive runs.
2. Return the (account, asset) pairs touched by the script, including those it would write to.

The hashing itself will be done in `ledger-v3-poc`; on the numscript side we only need to (a) emit store calls in a deterministic order and (b) expose touched accounts.

## What changed vs. the previous commit

- **No more full execution.** Previously `ResolveDependencies` was running phases 1-2 (vars + balance preload) and `RunProgram` was an alternative idea — both wasteful. It now performs phases 1-2 (records Reads via a `recordingStore` wrapper) then statically walks `Source` / `Destination` ASTs to collect Writes via `evaluateExprAs`. No posting is simulated.
- **Reads / Writes are separated in the output.** Writes are a conservative over-approximation: every account that appears as a source or destination is listed, even if a real run produces no posting for it. The `TestResolveDependencies_Nested` case now reports `{source1, source2, source3, world, dest}` instead of just `{source1, dest}` — which is what the host actually wants for preloading / locking.
- **Determinism.** Audited the call sites: `parseVars` iterates `[]VarDeclaration`, the AST traversal iterates `[]Statement` and `[]Source` slices, and `runBalancesQuery` is called once with a single map. The sequence of `GetBalances` / `GetAccountsMetadata` calls is therefore deterministic given identical inputs.
- **`ForbiddenFlags` is gone.** It only existed to reject `experimental-mid-script-function-call`, which was an issue when we did not execute statements. The static walker resolves every account expression via `evaluateExprAs`, which transparently handles mid-script `balance()` / `meta()` in source/destination positions, so the carve-out is no longer needed.

Re-flagging on your earlier point — the test now exercises a case where `s2 = meta(@account_that_needs_meta, "k")` resolves to `source2`. `Reads.Metadata` includes `account_that_needs_meta.k` and `Writes.Volumes` includes `source2` (statically discovered), so the host has everything it needs to load both into RAM before the hot path runs.

NumaryBot (commented):

The new API can miss real store reads and some actual balance writes for valid scripts, which breaks the dependency discovery and drift-detection guarantees it is meant to provide.

internal/interpreter/resolve_dependencies.go
```diff
@@ -0,0 +1,276 @@
...
+		if err := st.findBalancesQueriesInStatement(statement); err != nil {
+			return nil, err
+		}
+	}
+	if err := st.runBalancesQuery(); err != nil {
+		return nil, QueryBalanceError{WrappedError: err}
+	}
+
+	writes := map[string]map[string]struct{}{}
+	for _, statement := range program.Statements {
```
NumaryBot, line 115: 🔴 **[blocker] Include execution-time store reads**

For scripts where store-backed functions are evaluated during statement execution rather than source preloading, such as `set_tx_meta("k", meta(@cfg, "v"))` or caps/allotments using `balance(...)`, `Run` will call the store but this path only preloads source balances and then walks send statements without executing/evaluating those expressions. In those cases `ResolveDependencies` omits real metadata/balance reads from the admission hash, so FSM drift can go undetected.

internal/interpreter/resolve_dependencies.go
```diff
@@ -0,0 +1,276 @@
...
+	writes map[string]map[string]struct{},
+) InterpreterError {
+	switch source := source.(type) {
+	case *parser.SourceAccount:
+		return st.touchAccount(source.ValueExpr, source.Color, writes)
+
+	case *parser.SourceOverdraft:
+		return st.touchAccount(source.Address, source.Color, writes)
+
+	case *parser.SourceWithScaling:
```
NumaryBot, line 161: 🟠 **[major] Record scaling swap balance writes**

When a source uses `with scaling through @swap`, runtime emits conversion postings from the source to the swap account in scaled asset(s) and from the swap account back to the source before the final send. This records only the source/current asset, so callers relying on `Writes.Volumes` will miss the swap account and scaled source assets that are actually impacted for asset-scaling scripts.

internal/interpreter/resolve_dependencies.go
```diff
@@ -0,0 +1,273 @@
...
+
+	st.varOriginPosition = true
+	if program.Vars != nil {
+		if err := st.parseVars(program.Vars.Declarations, vars); err != nil {
+			return nil, err
+		}
+	}
+	st.varOriginPosition = false
+
+	for _, statement := range program.Statements {
```
NumaryBot, line 105: 🔴 **[blocker] Include reads from function and cap expressions**

For scripts where a store-reading expression appears outside variable declarations, sent values, or account address expressions — for example `set_tx_meta("x", meta(@cfg, "k"))` or a source/destination cap using `balance(@limits, USD/2)` — `Run` will evaluate those expressions and read the store, but this resolver only performs the balance preload pass here and never evaluates those statement/cap/allotment expressions. The returned `Reads` can therefore omit inputs that affect execution, so the admission/FSM drift hash can miss metadata or balance changes.

internal/interpreter/resolve_dependencies.go
```diff
@@ -0,0 +1,273 @@
...
+	writes *BalanceQuery,
+) InterpreterError {
+	switch source := source.(type) {
+	case *parser.SourceAccount:
+		return st.touchAccount(source.ValueExpr, source.Color, writes)
+
+	case *parser.SourceOverdraft:
+		return st.touchAccount(source.Address, source.Color, writes)
+
+	case *parser.SourceWithScaling:
```
NumaryBot, line 161: 🟠 **[major] Record scaling writes for swap and scaled assets**

When the script uses `source = @src with scaling through @swap`, the interpreter can emit postings from `src` in the consumed scaled asset rows and from `swap` in the target asset before the final send. This branch records only `src` with `CurrentAsset`, so `Writes.Volumes` misses balances such as `src/EUR` and `swap/EUR/2`; callers using writes for preloading or conflict detection will not protect balances the script can update.

NumaryBot (commented):

ResolveDependencies can omit concrete store reads that are evaluated by Run and can under-report writes for asset scaling. These gaps break the intended dependency discovery and drift/conflict detection semantics.

internal/interpreter/resolve_dependencies.go
```diff
@@ -0,0 +1,273 @@
...
+	// Volumes lists every (account, asset, color) tuple that may be impacted
+	// by a posting emitted by the script.
+	Volumes BalanceQuery
+}
+
+// ResolveDependenciesOptions configures ResolveDependencies behavior.
+type ResolveDependenciesOptions struct {
+	// FeatureFlags enables additional experimental features
+	// (same semantics as RunWithFeatureFlags).
+	FeatureFlags map[string]struct{}
```
ascandone, line 47: minor: I think we don't quite need to check feature flags at this point. They are just used to make sure the user intentionally enables experimental stuff. But we don't care about this in analysis. So I'd just remove this from the API, and allow every exp. feature.

Implementation-wise: either accept a nil map as a feature flag map within the state struct, or don't change the behaviour of that, and create a map with every feature flag by using the flags.AllFlags slice

internal/interpreter/resolve_dependencies.go
```diff
@@ -0,0 +1,273 @@
...
+	"github.com/formancehq/numscript/internal/parser"
+	"github.com/formancehq/numscript/internal/utils"
+)
+
+// ResolvedDependencies summarizes what a script reads from and writes to the
+// store. The caller can use it to preload data and to detect input drift
+// between successive runs.
+type ResolvedDependencies struct {
+	// Reads contains the data the script read from the store while resolving.
+	Reads ResolvedReads
```
ascandone, line 18: minor: a small optimization we could do could be to avoid registering value as "reads" if the value was previously set and not read. This is easy for meta(), and could be done to some extent with balances (but it's more complex to prove)

In pratice, it's probably a minor optimization (since I don't think it's common to use meta() like this), so we probably don't need to impl it now

ascandone: The interesting case would be with balances:

```
send [USD/2 100] (
  source = @world
  destination = @src
)

send [USD/2 50] (
  source = @src
  destination = @dest
)
```

here I think we don't need to actually pre-load src's balance (we already know it has at least 50cents). And those kind of situations may be more common that with meta(). But with balances is more complex to impl

PR #139, opened 2026-06-03
feat!: expose color as a first-class Posting property

gfyrag:
## Summary

Color is now carried as a dedicated field on the output `Posting` type and
as a separate dimension throughout the interpreter, instead of being
encoded as an asset suffix (`USD_RED`). The numscript ↔ host contract
becomes a clean two-dimensional `(asset, color)` model: downstream
consumers (ledgers, indexers, reporting) can segregate balances by
`(account, asset, color)` without parsing asset strings.

The semantics already exercised by the `experimental-asset-colors`
feature flag are unchanged. Only the wire format and Go API change.

## Breaking changes

| Surface | Before | After |
| --- | --- | --- |
| `Posting` | `{Source, Destination, Amount, Asset}` | + `Color string` (json `"color"`) |
| `BalanceQuery` | `map[string][]string` (asset strings, color encoded as suffix) | `map[string][]AssetColor` with `AssetColor{Asset, Color string}` |
| `Balances` (in-memory) | `map[account]map[asset]*big.Int` | `map[account]map[asset]map[color]*big.Int` |
| `Balances` (JSON wire) | flat `{"alice": {"USD/2": 100}}` | strict colored `{"alice": {"USD/2": {"": 100, "RED": 50}}}` |
| `coloredAsset()` helper | encoded `ASSET_COLOR` in asset strings | **removed** |
| `fundsQueue.Pull` signature | `Pull(amount, color *string)` | `Pull(amount)` — the color filter parameter was dead code, dropped |
| `fundsQueue.PullColored`, `PullUncolored` | unused wrappers | **removed** |

No backwards-compatibility shim. The flat-shorthand JSON shape
(`{"USD/2": 100}`) is **not** accepted — every balance amount must be
written under an explicit color, with `""` for the uncolored bucket. The
130 fixture `.num.specs.json` files in the repo were rewritten
accordingly in a single sweep (jq one-liner over the tree).

A small public helper `Uncolored(amount)` is added so building test
balances stays terse: `Uncolored(big.NewInt(100))` produces
`ColorBalance{"": amount}`.

## Why now

A consumer ledger (POC) needs strict color-segregated balances stored as
`(account, asset, color)` keys, with color flowing through the emitted
posting. As long as numscript strips color from postings and folds it
into the asset string, consumers cannot do that cleanly. This PR fixes
the contract upstream so we do not need to translate at the boundary.

## What's covered by tests

- **`color_semantics_test.go`** (black-box, 17 tests): color propagation
  through emitted postings, isolation between colors, source-restriction
  semantics, allocation distributing color across legs, send-all
  draining only the targeted color, JSON shape acceptance/rejection,
  charset enforcement (`^[A-Z]*$`).
- **`funds_queue_color_test.go`** (white-box, 1 test): `compactTop`
  never merges senders across colors. The `PullColored`/`PullUncolored`
  selectivity tests are gone with their subjects — the upstream
  invariant (`tryTakingFromAccount` caps push at the per-color balance,
  `tryTakingExact` validates completeness on pull) makes the filter
  parameter unnecessary.
- Updated `balances_test.go`, `interpreter_test.go`, `numscript_test.go`,
  `specs_format/*_test.go`, plus the 130 `.num.specs.json` fixtures.

## Test plan

- [x] All numscript test packages green; `just pre-commit` clean.
- [x] Downstream consumer (ledger POC) integrates against this branch
      via a `replace`/pseudo-version pin and exercises segregated
      balances end-to-end ([formancehq/ledger-v3-poc#234](https://github.com/formancehq/ledger-v3-poc/pull/234)).
- [x] CodeRabbit's 3 actionable comments addressed:
      - **Color + `with scaling`** (Critical): the grammar already
        refuses the combination, and the `SourceWithScaling.Color`
        field is never populated by the parser — leaving as-is.
      - **`HasPrefix` exact-match in `getAssets`** (Major): fixed in
        stacked PR [#140](https://github.com/formancehq/numscript/pull/140)
        on top of failing test [#141](https://github.com/formancehq/numscript/pull/141).
      - **MCP `float64 → big.Int` truncation** (Major): fixed in
        stacked PR [#143](https://github.com/formancehq/numscript/pull/143)
        on top of failing test [#142](https://github.com/formancehq/numscript/pull/142).


internal/mcp_impl/handlers.go
```diff
@@ -30,14 +30,25 @@ func parseBalancesJson(balancesRaw any) (interpreter.Balances, *mcp.CallToolResu
...
+			colorMap, ok := perAssetRaw.(map[string]any)
+			if !ok {
+				return interpreter.Balances{}, mcp.NewToolResultError(fmt.Sprintf("Expected {color: amount} object for %s/%s, got: <%#v>", account, asset, perAssetRaw))
+			}
+			for color, amountRaw := range colorMap {
+				amount, ok := amountRaw.(float64)
+				if !ok {
+					return interpreter.Balances{}, mcp.NewToolResultError(fmt.Sprintf("Expected float for amount: %v", amountRaw))
+				}
+				n, _ := big.NewFloat(amount).Int(new(big.Int))
```
flemzord: This still converts JSON numbers through `float64` and then `big.NewFloat(...).Int(...)`, which silently truncates fractions and can round large integers before conversion. Since balances are ledger amounts, this should reject non-integer or non-exact JSON numbers, or include the validation from PR #143 before #139 is merged standalone.

internal/interpreter/asset_scaling.go
```diff
@@ -38,13 +38,21 @@ func getAssetScale(asset string) (string, int64) {
...
+// Asset scaling operates on the uncolored bucket only — colored funds are not
+// implicitly converted across scales.
 func getAssets(balance AccountBalance, baseAsset string) map[int64]*big.Int {
 	result := make(map[int64]*big.Int)
-	for asset, amount := range balance {
-		if strings.HasPrefix(asset, baseAsset) {
-			_, scale := getAssetScale(asset)
-			result[scale] = amount
+	for asset, colorMap := range balance {
+		if !strings.HasPrefix(asset, baseAsset) {
```
flemzord, line 47: This matches any asset starting with `baseAsset`, so `getAssets(balance, "USD")` also includes assets like `USDC` or `USDX/2` in the scaling pool. The match should be exact base asset or `baseAsset + "/"`; otherwise unrelated asset balances can be converted during asset scaling.

internal/mcp_impl/handlers.go
```diff
@@ -30,14 +30,25 @@ func parseBalancesJson(balancesRaw any) (interpreter.Balances, *mcp.CallToolResu
...
-			if !ok {
-				return interpreter.Balances{}, mcp.NewToolResultError(fmt.Sprintf("Expected float for amount: %v", amountRaw))
+		for asset, perAssetRaw := range assets {
+			if iBalances[account][asset] == nil {
+				iBalances[account][asset] = interpreter.ColorBalance{}
 			}
 
-			n, _ := big.NewFloat(amount).Int(new(big.Int))
-			iBalances[account][asset] = n
+			// Expected shape: { "USD/2": { "": 100, "RED": 50 } }.
```
flemzord: The parser now expects the strict colored shape, but the MCP `evaluate` tool description still documents the old flat shape (`{"USD/2": 100}`). Callers following the advertised schema will get runtime errors. Please update the tool description/example to show `{"USD/2": {"": 100}}` and mention that `""` is the uncolored bucket.

internal/interpreter/interpreter.go
```diff
@@ -98,6 +119,7 @@ type Posting struct {
 	Destination string   `json:"destination"`
 	Amount      *big.Int `json:"amount"`
 	Asset       string   `json:"asset"`
+	Color       string   `json:"color"`
```
flemzord: Since `Color` is now part of the posting contract, pretty output should expose it too. `PrettyPrintPostings` still renders only source, destination, asset, and amount, so `numscript run --output-format pretty` and spec failure output can hide the only difference between two colored postings.

ascandone:
closing this, and moving the work here instead : https://github.com/formancehq/numscript/pull/160

PR #140, opened 2026-06-03
fix(asset_scaling): match base asset exactly, not as a string prefix

gfyrag:
## Summary

Tightens `getAssets` (in `internal/interpreter/asset_scaling.go`) so an entry
belongs to the `baseAsset` scaling pool **iff** its key is exactly
`baseAsset` or has the form `baseAsset/N`. Drops the spurious matches
the parent test PR pins.

This PR is stacked on **#141** (the failing regression test). Merging
this flips `TestGetAssetsRejectsSpuriousPrefixMatches` from red to
green.

## Stack

| PR | What it does | Target |
| --- | --- | --- |
| [#141](https://github.com/formancehq/numscript/pull/141) | Failing regression test pinning the bug | `main` |
| **this PR (#140)** | Tightens `getAssets` exact-match | #141 branch |

## Diff vs the parent test branch

Single 5-line change in `getAssets`:

```diff
- if strings.HasPrefix(asset, baseAsset) {
-     _, scale := getAssetScale(asset)
-     result[scale] = amount
- }
+ if asset != baseAsset && !strings.HasPrefix(asset, baseAsset+"/") {
+     continue
+ }
+ _, scale := getAssetScale(asset)
+ result[scale] = amount
```

Plus a doc-comment on the function.

## Test plan

- [x] `TestGetAssetsRejectsSpuriousPrefixMatches` passes (red on #141, green here).
- [x] All other tests stay green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #141, opened 2026-06-03
test(asset_scaling): pin getAssets prefix-match bug (failing)

gfyrag:
## Summary

Failing regression test demonstrating a pre-existing bug in
`internal/interpreter/asset_scaling.go::getAssets`. CI is **intentionally
red** on this branch — see the stacked fix PR below.

`getAssets(balance, baseAsset)` uses `strings.HasPrefix(asset, baseAsset)`
to gather scaled variants of an asset. The match is too loose:

| Stored key | Should match `USD`? | Currently matches? |
| --- | --- | --- |
| `USD` | yes | ✅ |
| `USD/2` | yes | ✅ |
| `USDT` | **no** | ❌ |
| `USDT/2` | **no** | ❌ |
| `USD_RED` (color-suffix-encoded) | **no** | ❌ |
| `USD_RED/2` | **no** | ❌ |

Because the result map is keyed by precision scale only, an entry
like `USD_RED/2 = 500` silently overwrites the true `USD/2 = 2` value
in the scaling pool — corrupting any subsequent conversion. The result
is non-deterministic (last-write-wins on randomized map iteration), so
the failing test is structured as three deterministic sub-tests rather
than one composite assertion.

## Stack

| --- | --- | --- |
| **this PR (#141)** | Adds the failing regression test | `main` |
| [#140](https://github.com/formancehq/numscript/pull/140) | Tightens `getAssets` to exact base + `base/N` match | this branch |

Merging the fix flips the test green.

## Test plan

- [x] `TestGetAssetsRejectsSpuriousPrefixMatches` fails locally with
      sub-test failures for `USDT` and `USD_RED` prefix pollution.
- [x] All other tests stay green.

### DIFF
--- a/internal/interpreter/asset_scaling_test.go
+++ b/internal/interpreter/asset_scaling_test.go
@@ -272,3 +272,49 @@ func TestUnboundedScalinHigherAssetTrimRemainder(t *testing.T) {
 		{3, big.NewInt(10)},
 	}, sol)
 }
+
+// getAssets must match a base asset exactly: assets that merely share the
+// base name as a string prefix (e.g. "USDT" or "USD_RED") are NOT scaled
+// forms of "USD" and must be ignored. Only the bare base and its
+// "base/N" precision variants belong in the scaling pool.
+//
+// Sub-tests are deliberately split so each is deterministic regardless of
+// Go's randomized map iteration order. A combined fixture is flaky:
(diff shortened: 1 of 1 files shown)


ascandone (approved):

ascandone:
Closing, as the relevant part was reworked and this bug no longer applies

PR #142, opened 2026-06-03
test(mcp_impl): pin parseBalancesJson float64 truncation bug (failing)

gfyrag:
## Summary

Failing regression tests demonstrating a silent corruption bug in
`internal/mcp_impl/handlers.go::parseBalancesJson`. CI is **intentionally
red** on this branch — see the stacked fix PR below.

JSON numeric values reach the MCP `evaluate` tool as `float64` (Go's
default when unmarshalling into `any`). The handler converts them with
`big.NewFloat(amount).Int(new(big.Int))`, which:

  - **Silently truncates fractional parts.** `100.9` becomes `100`.
  - **Silently rounds large values.** Anything past `±(2^53 - 1)` cannot be
    losslessly represented as a `float64`, so the integer the caller meant
    is gone before this handler ever sees it.

For a financial DSL these are not acceptable behaviors. The handler
should reject any value that isn't representable as an exact integer
in float64 precision.

## Stack

| --- | --- | --- |
| **this PR** | Adds failing regression tests | `main` |
| Fix PR (TBD) | Validates the amount before conversion | this branch |

Merging the fix flips the failing tests green.

## Test plan

- [x] `TestParseBalancesJsonRejectsNonIntegerAmounts` fails today (100.9 is silently truncated to 100).
- [x] `TestParseBalancesJsonRejectsUnsafelyLargeAmounts` fails today (1e18 is silently accepted).
- [x] `TestParseBalancesJsonAcceptsExactIntegerAmounts` passes (sanity).

### DIFF
--- a/internal/mcp_impl/handlers_test.go
+++ b/internal/mcp_impl/handlers_test.go
@@ -1,0 +1,66 @@
+package mcp_impl
+
+// Failing regression test demonstrating silent truncation in the MCP
+// `evaluate` tool's balance parser. JSON numeric values arrive in `any`
+// as `float64`, and the current implementation passes them straight
+// through `big.NewFloat(amount).Int(new(big.Int))` — which truncates
+// fractional parts and silently rounds values outside the exact-integer
+// range of float64 (±(2^53 - 1)).
+//
+// For a financial DSL, silent corruption of balance amounts is not
+// acceptable. The MCP handler should reject any value that cannot be
+// represented as an exact integer in float64 precision.
+//
+// CI is intentionally red on this branch. A stacked fix PR validates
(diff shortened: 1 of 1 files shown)


ascandone:
Closing — the function this pins (`parseBalancesJson`) has been removed; the evaluate tool now binds arguments into a typed `interpreter.Balances` via `BindArguments`.

That refactor does fix the fractional-amount case (`math/big.Int`'s `UnmarshalJSON` rejects non-integer literals), but not the large-magnitude case — the same class of bug still exists, just one layer up, in the transport's generic float64 decode. Opened #178 with a fix targeting the current code path, and closing this and #143 in favor of it.

— Claude

PR #143, opened 2026-06-03
fix(mcp_impl): validate JSON amounts as exact integers before big.Int conversion

gfyrag:
## Summary

Validates JSON amounts before converting to `*big.Int` in
`internal/mcp_impl/handlers.go::parseBalancesJson`. Stacked on **#142**
(the failing regression test); merging this flips the test green.

JSON numeric values reach the handler as `float64` (Go's default). The
previous code used `big.NewFloat(amount).Int(new(big.Int))`, which:

  - silently truncated fractional parts (`100.9` → `100`)
  - silently rounded values past `±(2^53 - 1)` (already lost in JSON
    decoding before reaching the handler)

The fix introduces a single guard before the conversion:

```go
if amount < -maxExactJSONInt || amount > maxExactJSONInt || amount != float64(int64(amount)) {
    return error
}
iBalances[account][asset] = big.NewInt(int64(amount))
```

where `maxExactJSONInt = 9_007_199_254_740_991` (`2^53 - 1`). The error
message names the offending `account/asset` so callers can pinpoint the
bad balance.

## Stack

| --- | --- | --- |
| [#142](https://github.com/formancehq/numscript/pull/142) | Failing regression tests | `main` |
| **this PR** | Adds the integer-validation guard | #142 branch |

## Test plan

- [x] `TestParseBalancesJsonRejectsNonIntegerAmounts` passes (red on #142, green here).
- [x] `TestParseBalancesJsonRejectsUnsafelyLargeAmounts` passes (red on #142, green here).
- [x] `TestParseBalancesJsonAcceptsExactIntegerAmounts` still passes.
- [x] All other tests stay green; `just pre-commit` clean.

### DIFF
--- a/internal/mcp_impl/handlers.go
+++ b/internal/mcp_impl/handlers.go
@@ -13,6 +13,11 @@ import (
 	"github.com/mark3labs/mcp-go/server"
 )
 
+// maxExactJSONInt is the largest integer that float64 can represent
+// without loss (2^53 - 1). JSON-decoded amounts past this magnitude have
+// already lost precision before reaching the handler, so we reject them.
+const maxExactJSONInt = float64(9_007_199_254_740_991)
+
 func parseBalancesJson(balancesRaw any) (interpreter.Balances, *mcp.CallToolResult) {
 	balances, ok := balancesRaw.(map[string]any)
 	if !ok {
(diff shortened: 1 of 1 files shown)


ascandone:
Closing — this fixes `parseBalancesJson`, which no longer exists (removed when the evaluate tool switched to `BindArguments` into a typed `interpreter.Balances`).

The underlying risk this PR addressed (silent precision loss on large balance amounts) is still real on the current code, just relocated to the MCP transport's generic JSON decode (`Params.Arguments any` always decodes numbers as float64, regardless of magnitude, before `BindArguments` ever runs). Opened #178 with the equivalent bounds-check applied at the current call site — closing this in favor of that.

— Claude

PR #144, opened 2026-06-04
fix: reject parse errors before execution

flemzord:
## Summary
- reject public `ParseResult.Run` calls when parsing produced errors
- return MCP evaluate parse errors immediately instead of continuing into execution
- add regression coverage for public Run and MCP evaluate

## TDD evidence
RED before fix:
- `go test . -run TestRunRejectsParseErrors -count=1 -v` panicked through `batch_balances_query.go` on an AST with nil source
- `go test ./internal/mcp_impl -run TestHandleEvalToolRejectsParseErrors -count=1 -v` failed before the handler seam/fix existed

GREEN after fix:
- `go test . -run TestRunRejectsParseErrors -count=1 -v`
- `go test ./internal/mcp_impl -run TestHandleEvalToolRejectsParseErrors -count=1 -v`
- `go test ./...`

### DIFF
--- a/internal/mcp_impl/handlers.go
+++ b/internal/mcp_impl/handlers.go
@@ -95,45 +95,47 @@ func addEvalTool(s *server.MCPServer) {
 			`),
 		),
 	)
-	s.AddTool(tool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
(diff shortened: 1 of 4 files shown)


ascandone (approved):
Thanks!

ascandone:
I'm merging that @flemzord

PR #145, opened 2026-06-04
fix: reject duplicate remaining allotments

flemzord:
## Summary
- reject allotment expressions with more than one `remaining` clause at runtime
- add public API regression coverage through `Parse(...).Run(...)`
- preserve existing supported runtime behavior where a single `remaining` can appear before a variable portion, as covered by `metadata.num`

## TDD evidence
RED before fix:
- `go test . -run TestRunRejectsDuplicateRemainingAllotments -count=1 -v` returned nil error for a source allotment with two `remaining` clauses

GREEN after fix:
- `go test . -run TestRunRejectsDuplicateRemainingAllotments -count=1 -v`
- `go test ./...`

## Notes
An initial broader runtime check for non-final `remaining` broke the existing `metadata.num` spec, which uses `remaining` before a variable commission portion. This PR therefore fixes the duplicate-remaining corruption path without narrowing the current runtime language semantics.

### DIFF
--- a/internal/interpreter/interpreter.go
+++ b/internal/interpreter/interpreter.go
@@ -1029,9 +1029,13 @@ func (s *programState) makeAllotment(monetary *big.Int, items []parser.Allotment
 			allotments = append(allotments, rat)
 
 		case *parser.RemainingAllotment:
+			if remainingAllotmentIndex != -1 {
+				return nil, InvalidRemainingAllotment{
+					Range: allotment.Range,
+				}
+			}
 			remainingAllotmentIndex = i
 			allotments = append(allotments, new(big.Rat))
-			// TODO check there are not duplicate remaining clause
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


ascandone (changes requested):
The general philosophy is that, with the new rewrite (well, not so "new" anymore), we separate static analysis and runtime errors, and reduce runtime errors to be more permissive and only block computations we can't possibly interpreter or succeed (unbound var, missing funds, etc). However, it's true that this one's actually confusing/misleading to allow, so it's probably a good idea to merge that.

Since we're here, I'd suggest to
1. emit a more precise error on the last (redudant) clause, not on the whole allotment, which is a bit noisy on e.g. editors (however that's debatable, one may also argue that error is best on the whole source). This one's optional
2. Avoiding writing the test using the flaky error message, but on the exact emitted struct. Also note most of the tests are on `interpreter_test` instead, whereas this test suite focuses more on the external contract. This one's more strongly suggested

ascandone (approved):

PR #152, opened 2026-06-11
fix: sort check diagnostics in source order with a valid comparator

flemzord:
## Summary

Fixes [EN-1120](https://formance-team.atlassian.net/browse/EN-1120).

The `numscript check` command sorted diagnostics with `sort.Slice` using `p2.GtEq(p1)` as the less function. This had two problems:

1. **Invalid comparator**: `Position.GtEq` is non-strict, so two diagnostics at the same position compare "less" both ways, violating the `sort.Slice` contract (undefined behavior).
2. **Reverse order**: the comparison sorted diagnostics descending, so they were printed in reverse source order.

## Changes

- Added a strict `Position.Lt` method in `internal/parser/range.go` (compares by line, then character), with unit tests.
- Extracted the sort in `internal/cmd/check.go` into a testable `sortDiagnostics` helper that sorts ascending using `Lt`.
- Added tests covering source-order sorting and equal-position safety.

## Testing

- `go test ./...` passes.
- `gofmt -l` clean on all touched files (the one flagged file, `internal/parser/parser.go`, is pre-existing on `main` and untouched here).
- Manual check on a file with two undeclared variables now prints diagnostics in ascending source order.

[EN-1120]: https://formance-team.atlassian.net/browse/EN-1120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/cmd/check.go
+++ b/internal/cmd/check.go
@@ -10,6 +10,15 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// sortDiagnostics sorts diagnostics in source order (by start position,
+// line first and then character), using a strict less function as required
+// by the sort.Slice contract.
+func sortDiagnostics(diagnostics []analysis.Diagnostic) {
+	sort.Slice(diagnostics, func(i, j int) bool {
+		return diagnostics[i].Range.Start.Lt(diagnostics[j].Range.Start)
+	})
+}
+
 func check(path string) {
 	dat, err := os.ReadFile(path)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


ascandone:
Closing, as the case this PR fix (order of errs with exactly the same range) is a non-deterministic implementation detail, thus won't be fixed after the PR either

PR #153, opened 2026-06-11
fix: don't panic on incomplete var declaration with origin

flemzord:
## Bug

`analysis.CheckSource` panics with a nil pointer dereference when checking a source containing an incomplete var declaration with an origin (`= ...`). Since there is no `recover` anywhere, this crashes the LSP server and `numscript check`.

Two nil derefs in `internal/analysis/check.go` (inside the `if varDecl.Origin != nil` branch of `check()`):

1. `*varDecl.Origin` can be a **nil `ValueExpr` interface**: when the origin expression is missing or malformed, `parseValueExpr` returns `nil`, but `parseVarDeclaration` still wrapped it in a non-nil `*ValueExpr` pointer. Calling `(*varDecl.Origin).GetRange()` in `unifyNodeWith` then panics.
2. `varDecl.Type` can be **nil** (e.g. `vars { $x = }`): nil guards exist elsewhere (lines 245 and 308) but `varDecl.Type.Name` was dereferenced unguarded.

## Reproduction

```go
analysis.CheckSource("vars { number $n = ")  // panic: nil pointer dereference
analysis.CheckSource("vars { number $n = }") // panic: nil pointer dereference
```

## Fix

- **Parser** (`internal/parser/parser.go`, `parseVarDeclaration`): don't produce a non-nil `Origin` pointer to a nil `ValueExpr` interface — leave `Origin` nil when the origin expression can't be parsed.
- **Analysis** (`internal/analysis/check.go`): defensively guard the origin check with `*varDecl.Origin != nil && varDecl.Type != nil`.
- Updated the `TestFaultToleranceIncompleteOrigin` snapshot, which had captured the buggy `Origin: &nil` value.
- Added regression tests in `internal/analysis/check_test.go` covering both repro inputs plus similar malformed variants, asserting `CheckSource` returns diagnostics without panicking.

`go test ./...` passes and `gofmt -l .` reports no unformatted files.

## References

Jira: [EN-1116](https://formance-team.atlassian.net/browse/EN-1116)

[EN-1116]: https://formance-team.atlassian.net/browse/EN-1116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/analysis/check.go
+++ b/internal/analysis/check.go
@@ -250,7 +250,7 @@ func (res *CheckResult) check() {
 				res.checkDuplicateVars(*varDecl.Name, varDecl)
 			}
 
-			if varDecl.Origin != nil {
+			if varDecl.Origin != nil && *varDecl.Origin != nil && varDecl.Type != nil {
 				res.checkExpression(*varDecl.Origin, varDecl.Type.Name)
 				res.unifyNodeWith(*varDecl.Origin, res.GetVarDeclType(varDecl))
 			}
--- a/internal/analysis/check_test.go
+++ b/internal/analysis/check_test.go
@@ -1193,3 +1193,26 @@ send [USD/2 100000] (
 	)
 
 }
+
+func TestIncompleteVarOriginDoesNotPanic(t *testing.T) {
+	t.Parallel()
+
+	// regression test for EN-1116: these incomplete var declarations
+	// used to make CheckSource panic with a nil pointer dereference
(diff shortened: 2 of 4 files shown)


ascandone (approved):
Thanks, merging

PR #154, opened 2026-06-11
fix: deep-copy balances returned by Store instead of mutating caller state

flemzord:
## Summary

Fixes [EN-1118](https://formance-team.atlassian.net/browse/EN-1118).

`Balances.Merge` stored the `*big.Int` pointers returned by `Store.GetBalances` directly into the interpreter's balance cache. The interpreter then mutates cached balances in place (`pushSender`, `pushReceiver`, and the `save` statement), so any third-party `Store` implementation returning pointers to its internal state had that state silently corrupted by `RunProgram`. The built-in stores happen to copy defensively, which is why the bug only affects external implementations.

## Changes

- `Balances.Merge` now deep-copies each amount (`new(big.Int).Set(amt)`) instead of aliasing the pointers returned by the `Store`.
- Added a doc comment on the `Store` interface clarifying that ownership of returned values transfers to the caller, and that implementations may safely return pointers to internal state.
- Added a regression test (`TestRunProgramDoesNotMutateStoreBalances`) using a `Store` stub that returns aliased `*big.Int`s; it verifies the stub's internal state is unchanged after a `send`. The test fails on `main` (balance drained from 100 to 0) and passes with this fix.

## Testing

- `go test ./...` passes.
- `gofmt -l .` only flags `internal/parser/parser.go`, which is pre-existing on `main` and untouched here.

[EN-1118]: https://formance-team.atlassian.net/browse/EN-1118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/interpreter/balances.go
+++ b/internal/interpreter/balances.go
@@ -69,6 +69,10 @@ func (b Balances) filterQuery(q BalanceQuery) BalanceQuery {
 }
 
 // Merge balances by adding balances in the "update" arg
+//
+// Values are deep-copied, so that the merged Balances never aliases the
+// *big.Int values owned by the "update" arg (which may belong to a
+// third-party Store implementation and must not be mutated in place)
 func (b Balances) Merge(update Balances) {
 	// merge queried balance
 	for acc, accBalances := range update {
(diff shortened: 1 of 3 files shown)


ascandone:
Closing, as this is implemented with outdated interfaces. Will fix in https://github.com/formancehq/numscript/pull/177 instead

PR #155, opened 2026-06-11
fix: test init feature-flag retry used zero default balance

flemzord:
## Summary

Fixes [EN-1123](https://formance-team.atlassian.net/browse/EN-1123).

In `makeSpecsFile` (`internal/cmd/test_init.go`), the missing-feature-flag retry branch passed `&missingFundsErr.Needed` to the recursive call. In that branch the `MissingFundsErr` type assertion had **failed**, so `missingFundsErr` was the zero value and the retry ran with a default balance of `0` instead of the current default (`100`). This caused generated specs to use wrong balances (e.g. the send amount instead of the intended default) whenever a script required a feature flag.

## Changes

- The feature-flag retry now passes the current `defaultBalance` parameter through instead of the zero-valued `&missingFundsErr.Needed`.
- `makeSpecsFile` recursion is now bounded by a `retriesLeft` counter (`maxSpecsRetries = 16`); exceeding it returns a clear error instead of recursing indefinitely.

## Tests

- Added `TestMakeSpecsFileFeatureFlagRetryKeepsDefaultBalance`: a script using `oneof` (which requires the `experimental-oneof` flag) with a send amount below the default balance now asserts the generated specs keep the `100` default balance. Verified the test fails (balance `42`) against the pre-fix code.
- `go test ./...` green, `gofmt -l` clean on changed files (`internal/parser/parser.go` is pre-existing on main).

[EN-1123]: https://formance-team.atlassian.net/browse/EN-1123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/cmd/test_init.go
+++ b/internal/cmd/test_init.go
@@ -104,15 +104,24 @@ func MakeSpecsFile(source string) (specs_format.Specs, error) {
 		vars,
 		map[string]struct{}{},
 		big.NewInt(100),
+		maxSpecsRetries,
 	)
 }
+
+// maxSpecsRetries bounds the number of times makeSpecsFile may retry
+// after recoverable interpreter errors (missing funds, missing feature flags).
+const maxSpecsRetries = 16
 
 func makeSpecsFile(
 	program parser.Program,
 	vars map[string]string,
 	featureFlags map[string]struct{},
 	defaultBalance *big.Int,
+	retriesLeft int,
 ) (specs_format.Specs, error) {
+	if retriesLeft < 0 {
+		return specs_format.Specs{}, fmt.Errorf("could not generate the specs file: exceeded the maximum number of retries (%d)", maxSpecsRetries)
+	}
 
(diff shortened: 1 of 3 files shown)


ascandone:
Closing, no follow-up needed: the bug this PR reports (feature-flag retry falling back to a zero balance instead of the caller-supplied default) is already fixed on main — `makeSpecsFile`'s feature-flag retry branch already passes `defaultBalance` through, not `&missingFundsErr.Needed`.

The PR also bundled an unrelated addition (a max-retry safeguard against unbounded recursion in `makeSpecsFile`). I didn't carry that forward: with a single shared default balance per retry, the recursion is bounded by the number of distinct accounts touched, so I don't see a live risk to guard against right now. Happy to reconsider if you'd like that filed as its own small PR.

— Claude

PR #156, opened 2026-06-11
fix: parse MCP balance amounts exactly instead of via float64

flemzord:
## Bug

Fixes [EN-1117](https://formance-team.atlassian.net/browse/EN-1117).

`parseBalancesJson` in `internal/mcp_impl/handlers.go` decoded balance amounts as `float64` and converted them via `big.NewFloat(amount).Int(...)`. This silently corrupted amounts:

- Integers at or above 2^53 were silently rounded (e.g. `9007199254740993` became `9007199254740992`).
- Fractional inputs like `100.5` were silently truncated to `100`.

For a monetary tool built on `big.Int`, silent precision loss is unacceptable.

## Fix

mcp-go (v0.45.0) unmarshals tool-call arguments with the standard `encoding/json` decoder into `map[string]any`, so JSON numbers always arrive as `float64` — the raw JSON is not recoverable at that point. The new `parseAmount` helper therefore:

- **Accepts string amounts** (and `json.Number`), parsed exactly into `big.Int` via `SetString` — this is the lossless path for arbitrarily large amounts.
- **Keeps backward compatibility for ordinary integer JSON numbers**: a `float64` is accepted only if it is an exact integer with `|n| < 2^53` (the float64 safe-integer range, where the decoded value is guaranteed to match the caller's literal).
- **Rejects everything else with a clear error** instead of truncating/rounding:
  - non-integer numbers (e.g. `100.5`)
  - integers at or above 2^53 sent as JSON numbers, with a message telling the caller to pass the amount as a string

The `balances` tool parameter description now documents the string form for large amounts.

## Tests

Added in `internal/mcp_impl/handlers_test.go`:

- exact small integers and negative amounts parse correctly
- `2^53 + 1` as a JSON string parses exactly
- `2^53 + 1` as a JSON number is rejected (not silently rounded), with the error suggesting the string form
- fractional amounts (`100.5`) are rejected
- non-numeric strings and invalid types are rejected
- `parseBalancesJson` round-trips a real JSON payload (mixed number/string amounts) exactly, and errors include account/asset context
- end-to-end `handleEvalTool` run moving `9007199254740993` units using a string balance

`go test ./...` and `gofmt -l` on the changed package are green (`internal/parser/parser.go` is flagged by gofmt but is pre-existing on `main` and untouched here).

[EN-1117]: https://formance-team.atlassian.net/browse/EN-1117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/mcp_impl/handlers.go
+++ b/internal/mcp_impl/handlers.go
@@ -2,8 +2,11 @@ package mcp_impl
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"math"
 	"math/big"
+	"strconv"
 	"strings"
 
 	"github.com/formancehq/numscript/internal/analysis"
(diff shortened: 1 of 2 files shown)


ascandone:
Closing, as it was fixed during the rework of other features

PR #158, opened 2026-06-11
fix: metadata cache merge and CsvPretty bounds check

flemzord:
Fixes [EN-1131](https://formance-team.atlassian.net/browse/EN-1131) — two minor robustness issues.

## meta() pseudo-cache (`internal/interpreter/function_exprs.go`)

Every `meta()` call queried the store and then **replaced** `s.CachedAccountsMeta` entirely instead of merging, so the "cache" was useless and previously fetched entries were clobbered.

- The cache is now checked first: a `(account, key)` pair that was already fetched no longer triggers a new store query.
- On a cache miss, the query result is merged into the cache via the existing `AccountsMetadata.Merge` instead of replacing it.
- The `MetadataNotFound` behavior is preserved: negative lookups are not cached, so a key absent from the store is queried again (and errors) on every call.

Covered by `TestMetaFunctionCachesQueries` and `TestMetaFunctionMissingKeyStillErrors` in `numscript_test.go`, using a store that (unlike `StaticStore`) only returns the queried pairs, making both the caching and the merge observable.

## CsvPretty ragged rows (`internal/utils/pretty_csv.go`)

`row[fieldIndex]` was unchecked and panicked whenever a row was shorter than the header (and `maxLengths[index]` panicked for rows longer than the header).

- Missing cells are now treated as empty strings, both for padding computation and rendering.
- Extra cells beyond the header length are ignored.

Covered by the `TestPrettyCsvRaggedRows` snapshot test.

## Checks

- `go test ./...` green
- `gofmt -l .` clean (only pre-existing hit is `internal/parser/parser.go`, already unformatted on main, untouched)

[EN-1131]: https://formance-team.atlassian.net/browse/EN-1131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/interpreter/function_exprs.go
+++ b/internal/interpreter/function_exprs.go
@@ -60,13 +60,23 @@ func meta(
 		return "", err
 	}
 
+	// Check the cache first, so that we don't query the store again
+	// for a (account, key) pair we already fetched
+	if value, ok := s.CachedAccountsMeta[*account][*key]; ok {
+		return value, nil
+	}
+
+	// Note: we do not cache negative lookups, so a key that is absent from
+	// the store is queried again (and errors again) on every meta() call
 	meta, fetchMetaErr := s.Store.GetAccountsMetadata(s.ctx, MetadataQuery{
 		*account: []string{*key},
 	})
 	if fetchMetaErr != nil {
 		return "", QueryMetadataError{WrappedError: fetchMetaErr}
 	}
-	s.CachedAccountsMeta = meta
+	// Merge the fetched metadata into the cache instead of replacing it,
+	// so that previously cached entries are preserved
+	s.CachedAccountsMeta.Merge(meta)
 
 	// body
 	accountMeta := s.CachedAccountsMeta[*account]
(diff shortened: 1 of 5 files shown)


ascandone:
Closing, no follow-up needed. This PR bundled two fixes:

- The metadata-cache-overwrite bug (`meta()` clobbering `CachedAccountsMeta` instead of merging) is already fixed on main — `evaluate_expr.go` now merges into the cache via `InternalAccountsMetadata.Merge`.
- The CsvPretty ragged-row panic is real on paper, but checking all 3 production callers (`Balances.PrettyPrint`, `AccountsMetadata.PrettyPrint`, `PrettyPrintPostings`), each one always builds every row as a fixed-length literal matching the header exactly — there's no path today that can actually produce a ragged row. Given that, adding bounds-checking now would be guarding against a scenario that can't currently happen.

Both halves are settled without needing this PR's diff. Worth revisiting the CsvPretty case if a future caller ever needs to pass variable-length rows.

— Claude

PR #160, opened 2026-06-12
feat: color as first class citizen

ascandone:
Main purpose of this PR is to have a clearer contract with the ledger in which:
* colors are an explicit field in the requested balances (and an explicit field in the balances returned from the ledger)
* colors are an explicit field in the output postings

In order to accomodate, that, we needed the also update the format used by specs format and inputs format, which is now a list of rows.

As a consequence, we needed to decouple the external, user facing interface to get inputs, and the data structure used internally (now called InternalBalances).
Other things were changed during the refactor, like a more broad use of the Value implementors (String, Asset, etc) to have a more "strongly typed" codebase (the conversion is zero cost, and defined types don't exist at runtime)

Pro tip for reviewers: close to the "filter files" ui, you can find a "filter by extension" and remove the .json. Maybe you already knew about that. I didn't


internal/interpreter/balances.go
```diff
@@ -2,110 +2,68 @@ package interpreter
...
-// given a BalanceQuery, return a new query which only contains needed (asset, account) pairs
-// (that is, the ones that aren't already cached)
-func (b Balances) filterQuery(q BalanceQuery) BalanceQuery {
-	filteredQuery := BalanceQuery{}
-	for accountName, queriedCurrencies := range q {
-		filteredCurrencies := utils.Filter(queriedCurrencies, func(currency string) bool {
-			return !b.has(accountName, currency)
-		})
+func (b Balances) PrettyPrint() string {
+	// TODO show colors
```
ascandone: TODO note to myself: don't forget about this

internal/interpreter/internal_balances.go
```diff
@@ -0,0 +1,111 @@
...
+func (b InternalBalances) Set(account string, asset string, color string, amount *big.Int) {
+	for i := range b[account] {
+		if b[account][i].Asset == asset && b[account][i].Color == color {
+			b[account][i].Amount = amount
+			return
+		}
+	}
+	b[account] = append(b[account], AccountBalance{
+		Asset:  asset,
+		Color:  color,
```
NumaryBot, line 80: 🟠 **[major] InternalBalances.Set stores *big.Int by pointer without cloning, risking aliasing mutations**
<sub>reported by claude</sub>

Set stores the passed-in `amount` pointer directly rather than cloning it. Merge() calls Set with the *big.Int taken directly from the BalanceRow returned by the Store. Because fetchBalance later mutates these values in-place (e.g. via balance.Sub/Add in pushSender, forcePushPostingUncolored, pushReceiver), the cache can alias and corrupt the slice owned by the Store. FromBalancesRows deliberately clones to avoid aliasing, making Set/Merge's omission of cloning an inconsistency that can silently corrupt caller-owned stores.

**Suggestion:** In Set, store `new(big.Int).Set(amount)` (guarding nil), or clone inside Merge before calling Set, to match the cloning contract established by FromBalancesRows and DeepClone.

internal/interpreter/interpreter.go
```diff
@@ -779,7 +807,9 @@ func (s *programState) tryTakingUpTo(source parser.Source, amount *big.Int) (*bi
 		}
```
NumaryBot, line 734: 🟠 **[major] Catch-all/wildcard balance queries cannot represent a queried-but-empty account, breaking asset scaling**
<sub>reported by gpt</sub>

For asset scaling, preload queries use wildcard patterns (e.g. `EUR/*`). StaticStore.GetBalances returns no rows when no stored asset matches such a query, and InternalBalances has no separate marker indicating the account was queried but had an empty result. The scaling path then checks `acc, ok := s.CachedBalances[string(account)]` and returns an InvalidUnboundedAddressInScalingAddress error when `ok` is false. An account with zero matching balances should behave as having no funds and produce missing-funds/fallback behavior, not an invalid-unbounded error. This also affects test-init: TestInitStore.GetBalances only defaults rows returned by StaticStore, so wildcard queries with no rows are never seeded on retry, potentially causing infinite retries.

**Suggestion:** Represent queried-but-empty accounts separately (e.g. a sentinel entry), or change the scaling code to treat a missing cache entry for non-`world` accounts as an empty balance slice. Also update TestInitStore/catch-all handling so wildcard queries can seed a usable default balance or otherwise terminate with a clear error instead of retrying indefinitely.

internal/interpreter/balances.go
```diff
@@ -2,110 +2,68 @@ package interpreter
...
+	}
+	return a.Cmp(b) == 0
 }
 
 func CompareBalances(b1 Balances, b2 Balances) bool {
-	return utils.Map2Cmp(b1, b2, func(ab1, ab2 *big.Int) bool {
-		return ab1.Cmp(ab2) == 0
-	})
+	if len(b1) != len(b2) {
+		return false
```
NumaryBot, line 85: 🟠 **[major] CompareBalances is incorrect when duplicate (account, asset, color) rows are present**
<sub>reported by claude, gpt</sub>

CompareBalances uses a length check plus findRow, which returns only the first matching row. If a Balances slice contains duplicate (account, asset, color) rows, different slices can compare as equal — e.g. [{a,EUR,"",1},{a,EUR,"",1}] incorrectly compares equal to [{a,EUR,"",1},{a,EUR,"",2}] because both expected rows match the single first actual row. The previous map-based representation could not have duplicate keys; the new slice representation has no such guarantee, so this can silently mask incorrect results in tests and specs.

**Suggestion:** Either validate and reject duplicate (account, asset, color) rows at construction/parse time, or implement comparison via a normalized map/multiset keyed by (account, asset, color) so all rows are accounted for correctly.

internal/interpreter/interpreter.go
```diff
@@ -624,7 +643,8 @@ func (s *programState) takeAll(source parser.Source) (*big.Int, InterpreterError
...
 
 		baseAsset, assetScale := getAssetScale(s.CurrentAsset)
-		acc, ok := s.CachedBalances[*account]
+		// TODO this may be a bug. Probably can't access cached balances directly this way. Double check
```
NumaryBot: 🟡 **[minor] Unresolved TODO comments flagging possible bugs in cached-balance access**
<sub>reported by claude</sub>

Two spots in takeAll/tryTakingUpTo access `s.CachedBalances[string(account)]` directly and carry comments like `// TODO this may be a bug. Probably can't access cached balances directly this way. Double check`. Shipping code the author explicitly flags as a possible bug is risky, especially in the asset-scaling path. Direct map access bypasses fetchBalance and returns whatever color rows happen to be cached. Additionally, getAssets now iterates AccountBalance entries of all colors, so it's unclear whether scaling is intentionally color-agnostic.

**Suggestion:** Resolve the TODOs: either confirm the access is correct (and remove the comment), or route through a proper accessor. Clarify whether scaling should be color-aware or color-agnostic.

internal/interpreter/asset_scaling.go
```diff
@@ -35,15 +36,15 @@ func getAssetScale(asset string) (string, int64) {
 		// fallback if parsing fails
 		return parts[0], 0
 	}
-	return asset, 0
+	return string(asset), 0
 }
```
NumaryBot: 🟡 **[minor] getAssets ignores color when matching scaled assets, potentially double-counting colored balances**
<sub>reported by claude</sub>

getAssets iterates accountBalances and matches purely on `strings.HasPrefix(accBalance.Asset, baseAsset)` with no color filter. Previously, color was fused into the asset string (e.g. COIN_RED), so colored balances would not have matched an uncolored base asset prefix. Now that color is a separate field, a colored entry (e.g. {Asset: EUR, Color: RED}) will be picked up by the scaling solver as if it were spendable uncolored EUR, potentially double-counting or incorrectly spending colored funds.

**Suggestion:** Filter accountBalances by the relevant color (likely `Color == ""`) when building the scale map, or explicitly document and test that scaling is intentionally color-agnostic.

internal/interpreter/balances.go
```diff
@@ -2,110 +2,68 @@ package interpreter
...
-// if the tuple is not present, it will write a big.NewInt(0) in it and return it
-func (b Balances) fetchBalance(account string, uncoloredAsset string, color string) *big.Int {
-	return utils.NestedMapGetOrPutDefault(b, account, coloredAsset(uncoloredAsset, &color), func() *big.Int {
-		return new(big.Int)
-	})
+type BalanceRow struct {
+	Account string   `json:"account"`
+	Asset   string   `json:"asset"`
+	Amount  *big.Int `json:"amount"`
+	Color   string   `json:"color,omitempty"`
```
NumaryBot, line 14: 🟡 **[minor] PrettyPrint omits color, making colored balance rows indistinguishable in output**
<sub>reported by gpt</sub>

PrettyPrint still renders only Account, Asset, and Balance, with a TODO to show colors. With color now a first-class field, colored and uncolored rows for the same account/asset are visually identical in this output, which makes debugging and test-failure output misleading.

**Suggestion:** Include a Color column in PrettyPrint output, or render the color alongside the asset, so colored balance rows are clearly distinguishable.

NumaryBot (commented):
## 🛑 Changes requested — multi-model review

This PR introduces color as a first-class field on postings and balances, replacing the previous convention of encoding color into the asset string (e.g. COIN_RED). The refactor is large but mostly mechanical, and the snapshot/spec regeneration provides good coverage. However, several correctness issues need to be addressed before merging.

The most critical concern is that the new slice-based balance comparison (CompareBalances) is incorrect in the presence of duplicate (account, asset, color) rows: findRow returns only the first match, so slices with different content but the same length can compare as equal, silently masking test failures. A related structural gap is that InternalBalances.Set stores *big.Int pointers directly without cloning, while FromBalancesRows and DeepClone do clone — this inconsistency can lead to the in-place arithmetic in fetchBalance corrupting memory owned by the caller's Store.

For asset scaling, catch-all/wildcard queries that return no rows leave no trace in InternalBalances, causing the cache-miss path to return an invalid-unbounded error instead of treating the account as having zero funds. This breaks the scaling fallback and may also cause test-init to retry indefinitely for wildcard queries. Additionally, getAssets matches purely on asset-name prefix with no color filter, meaning colored entries (e.g. EUR/RED) are now inadvertently included in scaling calculations that previously relied on color being baked into the asset string.

Unresolved TODO comments in takeAll/tryTakingUpTo explicitly note that the direct cache access "may be a bug," which should be resolved before shipping. PrettyPrint still omits the color field (noted with a TODO), making colored and uncolored rows for the same account/asset visually indistinguishable in debug output. A minor typo ("overdrat") appears copy-pasted in two comment sites. The missing-funds retry convergence for multi-source scripts should also be validated with a test.

### Findings outside the diff

🟡 **[minor] Missing-funds retry may not converge for multi-source scripts** — `internal/cmd/test_init.go:138` <sub>(claude)</sub>

On retry, makeSpecsFile raises DefaultBalance to Needed. For a single-account script this converges, but for multi-source scripts it is not verified that Needed decreases relative to the updated DefaultBalance on each recursive call. If it doesn't, the retry could recurse unboundedly. The test only covers a single-account case.

**Suggestion:** Add a regression test for a multi-source missing-funds script to confirm the retry converges, and verify that the per-row funding logic interacts correctly with the single global DefaultBalance.

---
<sub>Reviewed in parallel by **claude** (anthropic/claude-opus-4-8) and **gpt** (openai/gpt-5.5), then consolidated. A new review is posted on each push.</sub>

internal/interpreter/function_exprs.go
```diff
@@ -23,26 +23,27 @@ func overdraft(
 	asset := parseArg(p, r, expectAsset)
 	err = p.parse()
 	if err != nil {
-		return nil, err
+		return Monetary{}, err
```
NumaryBot, line 26: ⚪ **[nit] Typo in comment: 'overdrat' (copy-pasted twice)**
<sub>reported by claude</sub>

Comment reads `// overdrat call doesn't handle colors` in both overdraft() and balance(). This typo is copy-pasted in two places.

**Suggestion:** Fix to 'overdraft' and ensure the comment in balance() describes the balance() function rather than overdraft.

internal/interpreter/interpreter.go
NumaryBot: 🟠 **[major] Direct CachedBalances map access for scaling source may fail or silently use stale data**
<sub>reported by claude, gpt</sub>

Both `takeAll` (~line 645) and `tryTakingUpTo` (~line 810) access `s.CachedBalances[string(account)]` directly and bail with `InvalidUnboundedAddressInScalingAddress` when the account key is absent. With the new slice-based `InternalBalances` cache, an entry is only present if it was previously fetched. With the old map representation, a catch-all query like `EUR/*` materialized an empty account entry even for zero matches, so scaling could proceed and ultimately report missing funds rather than an invalid-address error. The new `StaticStore.GetBalances` returns no rows for an empty catch-all, and `InternalBalances.Merge` therefore never creates the account key. This changes zero/empty scaling sources from normal insufficient-funds behavior into an invalid-address error. The author left 'TODO this may be a bug' and 'TODO double check access here. May be a bug.' comments in both locations, indicating these paths are unverified.

**Suggestion:** Do not use account-key existence as the signal for invalid address. Treat a missing account entry as an empty balance list and let the normal scaling/missing-funds path run, or teach `InternalBalances` to record empty catch-all query results as fetched markers. Resolve both TODOs and add regression tests covering scaling sources whose accounts aren't otherwise queried, including zero-balance catch-all cases.

internal/interpreter/interpreter.go
```diff
@@ -692,15 +713,20 @@ func (s *programState) takeAll(source parser.Source) (*big.Int, InterpreterError
 }
 
 // Fails if it doesn't manage to pull exactly "amount"
```
NumaryBot, line 643: 🟠 **[major] Negative-amount validation relocated, changing when NegativeAmountErr is raised**
<sub>reported by claude</sub>

Previously `runSendStatement` checked `monetaryAmt.Cmp(0) == -1` and returned `NegativeAmountErr` before calling `tryTakingExact`. The check now lives inside `tryTakingExact`. Additionally, `tryTakingUpTo` now unconditionally calls `utils.NonNeg(amount)`, which silently clamps negative amounts rather than erroring. This means send-all and allotment paths may no longer surface a negative top-level amount the same way, and the `NonNeg` clamping could mask negative amounts that previously produced an explicit error. This is a behavioral change with no corresponding test coverage.

**Suggestion:** Confirm that all paths which previously produced `NegativeAmountErr` still do (e.g. `send [USD/2 -1]`). Ensure `NonNeg` clamping in `tryTakingUpTo` does not silently swallow amounts that should error. Add or preserve an explicit test for a negative literal send amount.

internal/interpreter/asset_scaling.go
```diff
@@ -35,15 +36,15 @@ func getAssetScale(asset string) (string, int64) {
...
-func getAssets(balance AccountBalance, baseAsset string) map[int64]*big.Int {
+func getAssets(accountBalances []AccountBalance, baseAsset string) map[int64]*big.Int {
 	result := make(map[int64]*big.Int)
-	for asset, amount := range balance {
-		if strings.HasPrefix(asset, baseAsset) {
-			_, scale := getAssetScale(asset)
-			result[scale] = amount
+	for _, accBalance := range accountBalances {
+		if strings.HasPrefix(accBalance.Asset, baseAsset) {
+			_, scale := getAssetScale(Asset(accBalance.Asset))
```
NumaryBot: 🟠 **[major] Asset scaling ignores balance color when selecting convertible balances**
<sub>reported by gpt</sub>

`getAssets` receives all cached balances for an account and filters only by asset prefix. Since `AccountBalance` now has a separate `Color` field, this can mix colored and uncolored balances if the cache already contains rows for multiple colors. For example, an uncolored scaling source can consume a cached `RED` balance, or a colored scaling source can consume uncolored funds, producing postings that violate color isolation. The surrounding code already queries catch-all balances with a color, but this helper discards that dimension when reading from the account cache.

**Suggestion:** Pass the requested color into `getAssets` and filter on both asset family and `accBalance.Color`. Add tests combining `experimental-asset-scaling` with colored balances and colored/uncolored sources.

internal/interpreter/balances.go
```diff
@@ -2,110 +2,82 @@ package interpreter
...
+	return a.Cmp(b) == 0
 }
 
 func CompareBalances(b1 Balances, b2 Balances) bool {
-	return utils.Map2Cmp(b1, b2, func(ab1, ab2 *big.Int) bool {
-		return ab1.Cmp(ab2) == 0
-	})
+	if len(b1) != len(b2) {
+		return false
+	}
```
NumaryBot, line 86: 🟡 **[minor] CompareBalances is incorrect when duplicate (account, asset, color) rows exist**
<sub>reported by claude, gpt</sub>

The `Balances` type is now a slice, making duplicate `(account, asset, color)` rows possible. `CompareBalances` only checks slice length equality plus `CompareBalancesIncluding`; `CompareBalancesIncluding` uses `findRow`, which returns the first matching row. Two unequal balance sets with duplicate keys can therefore compare as equal — for example, expected rows `[A=5, A=5]` and actual rows `[A=5, A=7]` have equal length and both expected rows match the first actual row. This can cause tests and spec checks to pass when balances actually differ. The old map-based representation made duplicate keys impossible.

**Suggestion:** Either guarantee uniqueness of `(account, asset, color)` on construction (e.g. reject or merge duplicates in `findRow`/insert), or make `CompareBalances` a true multiset/bijection comparison rather than 'subset + equal length'. Add a test with duplicate keys to document the intended semantics.

internal/interpreter/internal_balances.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		cloned[account] = clonedEntries
+	}
+	return cloned
+}
+
+// Get the (account, asset, color) balance from the cache.
+// If it is not present, it writes a zero balance in it and returns it.
+func (b InternalBalances) fetchBalance(account AccountAddress, asset Asset, color String) *big.Int {
+	acc := string(account)
```
NumaryBot, line 53: 🟡 **[minor] O(n) linear scans replace O(1) map lookups in hot balance paths**
<sub>reported by claude</sub>

`fetchBalance`, `Set`, and `has` now linearly scan the per-account slice on every call. `fetchBalance` is invoked very frequently during interpretation (pushSender, pushReceiver, forcePushPostingUncolored, takeAll/tryTaking). For accounts with many `(asset, color)` entries this is quadratic compared to the previous map-based `AccountBalance`. The doc comment even acknowledges this representation 'could change over time, for example to add more indexes for faster lookups' but currently has none.

**Suggestion:** If accounts can hold many asset/color pairs, consider keeping an inner map keyed by `(asset, color)` for O(1) lookups, or benchmark to confirm the linear scan is acceptable for expected workloads.

internal/interpreter/internal_balances.go
```diff
@@ -0,0 +1,111 @@
...
+		Color:  string(color),
+		Amount: amount,
+	})
+	return amount
+}
+
+// Set assigns amount to the (account, asset, color) balance.
+func (b InternalBalances) Set(account string, asset string, color string, amount *big.Int) {
+	for i := range b[account] {
+		if b[account][i].Asset == asset && b[account][i].Color == color {
```
NumaryBot, line 73: 🟡 **[minor] InternalBalances.Set stores *big.Int pointer directly, risking aliasing with caller-owned values**
<sub>reported by claude</sub>

`Set` stores the passed `amount *big.Int` pointer directly (`b[account][i].Amount = amount` and in the append path). `Merge` passes `row.Amount` straight through. Since `Merge` receives `BalanceRow` values from `Store.GetBalances`, the cache may alias the same `*big.Int` the store returned. Subsequent in-place mutations via `fetchBalance().Sub`/`Add` will then mutate that shared pointer. `StaticStore.GetBalances` currently allocates with `new(big.Int).Set(...)` so it is currently safe, but this is an implicit contract that any Store returning shared or cached `*big.Int` pointers would silently break.

**Suggestion:** Clone the `*big.Int` in `Set` (or in `Merge`) before storing, mirroring `FromBalancesRows` which already clones, to avoid aliasing store-owned values.

internal/interpreter/interpreter.go
```diff
@@ -39,44 +42,45 @@ type StaticStore struct {
...
-					if !matchesAsset {
-						continue
-					}
-					outputAccountBalance[k] = new(big.Int).Set(v)
+	var output Balances
+	for _, item := range q {
+		baseAsset, isCatchAll := strings.CutSuffix(item.Asset, "/*")
+
+		if isCatchAll {
+			// return every stored asset (of the queried color) under the base asset
```
NumaryBot: 🟡 **[minor] StaticStore.GetBalances is O(query × balances) due to nested linear scan**
<sub>reported by claude</sub>

For each query item the function scans the entire `s.Balances` slice (for both catch-all and exact lookup). With the previous map representation lookups were O(1). For large balance sets and many query items this becomes O(q×n). `GetBalances` is on the preload critical path.

**Suggestion:** Build a one-time index (map keyed by account, or account+asset+color) at the start of `GetBalances` if the balance set is expected to be large; otherwise document that `StaticStore` is intended for small/test inputs only.

internal/interpreter/function_exprs.go
```diff
@@ -23,26 +23,27 @@ func overdraft(
 	asset := parseArg(p, r, expectAsset)
 	err = p.parse()
 	if err != nil {
```
NumaryBot, line 25: ⚪ **[nit] Misleading copy-pasted and misspelled comment in balance()**
<sub>reported by claude</sub>

Both `overdraft()` and `balance()` carry the comment `// overdrat call doesn't handle colors`. In `balance()` the reference to 'overdrat call' is a copy-paste leftover and is also misspelled, making the intent unclear.

**Suggestion:** Fix to e.g. `// the balance() builtin doesn't handle colors` and correct the spelling in both locations.

NumaryBot (commented):
## 🛑 Changes requested — multi-model review

This PR refactors the balance representation from a nested map to a flat slice of `BalanceRow`, promotes asset color to a first-class column, and introduces a separate `InternalBalances` cache for the runtime. The structural direction is sound and the mechanical migrations are consistent, but several correctness issues need to be resolved before merging.

The most critical problems are: (1) Two locations in the scaling source paths (`takeAll` and `tryTakingUpTo`) access the `CachedBalances` map directly by account key and return `InvalidUnboundedAddressInScalingAddress` when the key is absent — the author flagged both with TODO comments. With the new lazy-populated cache, empty catch-all queries no longer materialize an account entry, so valid-but-empty scaling sources now produce a wrong error instead of the expected insufficient-funds path. (2) The negative-amount validation was moved inside `tryTakingExact`, and `tryTakingUpTo` now silently clamps negative values via `NonNeg` — this is a behavioral change that needs explicit test coverage to confirm `NegativeAmountErr` is still raised on all previously-erroring paths. (3) The asset scaling helper `getAssets` filters cached balances by asset prefix only, ignoring the new `Color` field, so colored and uncolored balances can be mixed during scaling operations in violation of color isolation.

Additional minor concerns: `CompareBalances` is no longer correct when duplicate `(account, asset, color)` rows exist in the slice (both reviewers flagged this); `InternalBalances.Set` stores `*big.Int` pointers directly without cloning, creating an implicit aliasing contract with stores; hot paths degrade from O(1) map lookups to O(n) linear scans; and there is a copy-pasted misspelled comment and an unkeyed struct literal in tests.

### Findings outside the diff

⚪ **[nit] StaticStore initialized with positional (unkeyed) struct fields** — `internal/interpreter/interpreter_test.go:142` <sub>(claude)</sub>

`machine.StaticStore{testCase.balances, testCase.meta}` uses unkeyed struct literals. If `StaticStore` fields are ever reordered this silently breaks. This PR is actively reshaping these types, making the risk higher.

**Suggestion:** Use keyed fields: `machine.StaticStore{Balances: testCase.balances, Meta: testCase.meta}`.

---
<sub>Reviewed in parallel by **claude** (anthropic/claude-opus-4-8) and **gpt** (openai/gpt-5.5), then consolidated. A new review is posted on each push.</sub>

internal/interpreter/interpreter.go
ascandone: The static store is mostly used for testing so we don't really care about that. Besides, we're talking about sizes of data for which linear scan is most likely faster than hashmaps

internal/interpreter/interpreter.go
```diff
@@ -383,41 +314,58 @@ type programState struct {
...
+	amtBi := big.Int(amount)
+
+	if amtBi.Sign() == 0 {
 		return
 	}
 
-	srcBalance := st.CachedBalances.fetchBalance(posting.Source, posting.Asset, "")
-	srcBalance.Sub(srcBalance, posting.Amount)
+	srcBalance := st.CachedBalances.fetchBalance(source, asset, "")
+	srcBalance.Sub(srcBalance, &amtBi)
```
NumaryBot, line 357: 🟠 **[major] Asset scaling ignores color when creating swap postings**
<sub>reported by gpt</sub>

`forcePushPostingUncolored` always fetches and updates balances with color `""` and emits postings with no color. The `SourceWithScaling` paths call this helper for conversions even when `source.Color` is set. As a result, scaling a colored balance debits an uncolored `(account, asset, "")` row, credits an uncolored scaled asset back, and then `tryTakingFromAccount(..., source.Color)` still looks at the colored target balance. Colored funds therefore cannot be scaled correctly and may produce missing funds or incorrect internal balances/postings.

**Suggestion:** Make the scaling swap path color-aware: pass the evaluated source color into the conversion helper, update `(asset, color)` balances, and emit postings with the same color. Alternatively, explicitly reject `SourceWithScaling` with a color if colored scaling is unsupported and document that limitation.

NumaryBot (commented):
## 🛑 Changes requested — multi-model review

This PR refactors balances from a nested-map representation to a flat row-based `(account, asset, color)` structure, making asset color a first-class concept. The design direction is sound and the regenerated test data appears stable, but several correctness issues need to be addressed before merging. The most critical are: asset scaling ignoring color entirely when creating swap postings (causing incorrect internal bookkeeping for colored balances), empty catch-all scaling queries being misclassified as invalid-address errors, potential aliasing of `*big.Int` pointers between postings and cached balances, and a rangeless `NegativeAmountErr` from `tryTakingExact` that degrades user-facing diagnostics. There are also self-acknowledged TODO comments marking possible bugs in core scaling paths, inconsistencies in the `InternalBalances.Set` mutation model, and minor issues around typos and redundant conversions.

### Findings outside the diff

🟠 **[major] Empty catch-all scaling queries treated as invalid accounts instead of empty balances** — `internal/interpreter/interpreter.go:579` <sub>(claude, gpt)</sub>

For scaled sources, the runtime checks `acc, ok := s.CachedBalances[string(account)]` and returns `InvalidUnboundedAddressInScalingAddress` when no entry exists. With the new row-based `StaticStore.GetBalances`, a catch-all query (e.g. `EUR/*`) returns no rows when the account has no matching balances, leaving no account key in the cache. This turns an empty balance into an invalid-address error before the normal missing-funds path can run, which is a behavioral regression. This is also flagged by an explicit TODO comment in the code acknowledging it may be a bug.

**Suggestion:** Treat a missing account entry in the cache after a completed catch-all balance query as an empty slice in the scaling code rather than an error. Add explicit cache state for completed catch-all queries or known empty accounts, then let the existing withdrawal logic produce the appropriate missing-funds result.

🟠 **[major] Postings and cached balances may alias the same *big.Int pointer** — `internal/interpreter/interpreter.go` <sub>(claude)</sub>

In `forcePushPostingUncolored`, `amtBi := big.Int(amount)` is taken once and `&amtBi` is used both for in-place balance arithmetic (Sub/Add) and stored directly in the `Posting.Amount` field. If any later code mutates a cached balance that shares this pointer, or mutates the posting amount, they can interfere. In `pushReceiver`, `postings.Amount = sender.Amount` is reused for `destBalance.Add`, making the pattern fragile. Previously a freshly created `*big.Int` was passed for each posting.

**Suggestion:** Store a fresh copy in the Posting: `Amount: new(big.Int).Set(&amtBi)`, ensuring postings never share mutable state with internal balance bookkeeping.

🟠 **[major] NegativeAmountErr from tryTakingExact lacks a source Range** — `internal/interpreter/interpreter.go` <sub>(claude)</sub>

The negative-amount check was moved from `runSendStatement` (where a Range was set) into `tryTakingExact`, where the returned `NegativeAmountErr{Amount: amount}` has no Range attached. The `interpreter_test` `removeRange` helper masks this in tests. This degrades user-facing diagnostics because error locations are lost. Additionally, moving the guard changes which errors surface for negative monetary literals feeding allotment sources, and the exact behavior may not be covered by existing tests.

**Suggestion:** Plumb the source range into `tryTakingExact` (it already calls `source.GetRange()` for `MissingFundsErr`) and set it on `NegativeAmountErr`. Add or verify a test asserting `NegativeAmountErr` for `send [USD/2 -1] (...)` still reports the expected source range.

🟡 **[minor] Self-acknowledged possible bug in asset-scaling cached-balance access (TODO comments)** — `internal/interpreter/interpreter.go` <sub>(claude)</sub>

Two TODO comments read: `// TODO this may be a bug. Probably can't access cached balances directly this way. Double check` around `acc, ok := s.CachedBalances[string(account)]`. With the new `InternalBalances` representation, indexing the map returns the cached slice only if it was populated; for scaling sources balances are intentionally not queried, so `ok` may behave differently than with the old map. Shipping code with explicit 'this may be a bug' TODOs in core paths is risky.

**Suggestion:** Resolve the TODOs before merge: confirm with a test that scaling sources still raise `InvalidUnboundedAddressInScalingAddress` when expected and read the correct cached entries, or fix the access pattern.

🟡 **[minor] InternalBalances.Set replaces pointer instead of mutating in place, inconsistent with fetchBalance** — `internal/interpreter/internal_balances.go` <sub>(claude)</sub>

`Set` does `b[account][i].Amount = amount` (pointer replacement), whereas `fetchBalance` returns the existing `*big.Int` that callers mutate in place via Sub/Add. After a `Merge` (which calls `Set` with the queried row's `*big.Int`), a previously fetched pointer for the same `(account, asset, color)` would become stale/detached from the cache. In practice `Merge` happens during `runBalancesQuery` before `fetchBalance` is used in arithmetic, so it is likely safe today, but the inconsistency is a latent bug if call ordering changes.

**Suggestion:** Make `Set` copy into the existing slot's `*big.Int` (`b[account][i].Amount.Set(amount)`) or document clearly that `Set` must only be called before any `fetchBalance` pointers are taken.

🟡 **[minor] StaticStore catch-all now filters by exact color, potentially hiding colored balances from scaling** — `internal/interpreter/store.go` <sub>(claude)</sub>

The new `GetBalances` catch-all branch only returns rows where `row.Color == item.Color`. For a catch-all query with an empty color, colored rows (e.g. color `RED`) are excluded. Combined with asset-scaling (which uses `assetToScaledAsset => BASE/*`), this means scaled balances with a non-empty color are silently invisible to scaling. This interaction is not clearly tested or documented.

**Suggestion:** Add a test for a catch-all/scaling query against an account holding colored sub-assets, or explicitly document that scaling ignores colored balances by design.

🟡 **[minor] Possible non-determinism in spec generation when GetBalances appends during iteration** — `internal/cmd/test_init.go` <sub>(claude)</sub>

`TestInitStore.GetBalances` appends newly-funded rows to `s.StaticStore.Balances` during the same call. The `stored` set prevents duplicates within a call, but if a retry re-queries after funding, the order of generated balances depends on map/slice iteration and append order, potentially producing non-deterministic spec files across runs.

**Suggestion:** Confirm specs generation is deterministic (e.g. sort balances before writing the `.num.specs.json` file) to avoid noisy diffs in generated testdata.

⚪ **[nit] Misspelled and misleading comment 'overdrat call doesn't handle colors' in balance()** — `internal/interpreter/function_exprs.go` <sub>(claude)</sub>

The comment `// overdrat call doesn't handle colors` in the `balance()` function is both misspelled ('overdrat') and incorrect — it refers to overdraft, not balance. This is a copy-paste artifact that will confuse future readers.

**Suggestion:** Fix the typo: change 'overdrat' to 'overdraft' in `overdraft()`, and reword the comment in `balance()` to something like `// balance() does not support colors yet`.

⚪ **[nit] Redundant Asset() conversion in balance()** — `internal/interpreter/function_exprs.go` <sub>(claude)</sub>

In `balance()`, `Asset: Asset(asset)` is used where `asset` is already of type `Asset` (returned from `parseArg` with `expectAsset`). The conversion is a no-op and is inconsistent with `overdraft()` which uses `Asset: asset` directly.

**Suggestion:** Drop the redundant `Asset(...)` conversion for clarity and consistency with `overdraft()`.

---
<sub>Reviewed in parallel by **claude** (anthropic/claude-opus-4-8) and **gpt** (openai/gpt-5.5), then consolidated. A new review is posted on each push.</sub>

ascandone:
@corerabbitai review the whole thing again (discard stale comments)

ascandone:
@coderabbitai @NumaryBot please review the whole PR again. Discard the outdated comments, and make sure you take new ones into account.
Note:
1. scaling doesn't allow colors. Implementing that is out of scope for now.
2. if you find errors, try to reproduce them (at least informally in the comment), with an actual reproduction using numscript, with the precondition (e.g. balances or meta), expected result and actual result (postings/meta)

internal/mcp_impl/handlers.go
```diff
@@ -113,22 +61,21 @@ func handleEvalTool(ctx context.Context, request mcp.CallToolRequest) (*mcp.Call
 		return mcp.NewToolResultError(strings.Join(out, ", ")), nil
 	}
 
-	balances, mcpErr := parseBalancesJson(request.GetArguments()["balances"])
-	if mcpErr != nil {
-		return mcpErr, nil
+	var args struct {
```
NumaryBot, line 66: 🟠 **[major] Accept the new balance shape in the MCP tool**

With `interpreter.Balances` now being `[]BalanceRow`, MCP callers that follow the advertised `balances` object schema/description still send `{ "alice": { "USD": 10 } }`; `BindArguments` cannot unmarshal that object into a slice, so the `evaluate` tool fails before running any script with the documented input. Update the tool schema/description to an array or keep translating the legacy nested map here.

internal/cmd/test_init.go
```diff
@@ -200,40 +201,52 @@ func runTestInitCmd(opts testInitArgs) error {
...
 
-func (s TestInitStore) GetBalances(_ context.Context, q interpreter.BalanceQuery) (interpreter.Balances, error) {
-	outputBalance := interpreter.Balances{}
-	for queriedAccount, queriedCurrencies := range q {
-
-		for _, curr := range queriedCurrencies {
-			amt := utils.NestedMapGetOrPutDefault(s.Balances, queriedAccount, curr, func() *big.Int {
-				return new(big.Int).Set(s.DefaultBalance)
-			})
+func (s *TestInitStore) GetBalances(ctx context.Context, q interpreter.BalanceQuery) (interpreter.Balances, error) {
```
NumaryBot, line 207: 🟠 **[major] Seed default balances for scaling wildcard queries**

For `test-init` on an asset-scaling script, the preload query asks for a wildcard asset like `USD/*`. Starting from an empty store, `StaticStore.GetBalances` returns no rows for that catch-all query, so this wrapper never creates a default balance and the interpreter later raises `InvalidUnboundedAddressInScalingAddress` instead of generating a specs file. The previous store funded the queried wildcard, so this regresses `test-init` for scaling sources without preexisting balances.

internal/interpreter/interpreter.go
```diff
@@ -98,6 +28,7 @@ type Posting struct {
 	Destination string   `json:"destination"`
 	Amount      *big.Int `json:"amount"`
 	Asset       string   `json:"asset"`
+	Color       string   `json:"color,omitempty"`
```
NumaryBot, line 31: 🟠 **[major] Include posting colors in pretty output**

Now that postings can carry a non-empty `Color`, the CLI/spec runner pretty-printer still renders only source, destination, asset, and amount. For colored sends, `numscript run` and failure output lose the color column, making distinct colored movements look identical even though the JSON result contains the color.

NumaryBot (commented):

The core color model is partially wired, but the MCP evaluate tool still exposes the old balance input shape, test-init regresses for scaling wildcard balance queries, and colored postings are omitted from human-readable output.

ascandone:
@coderabbitai
> Bug 1 — StaticStore.GetBalances panics when row.Amount is nil
false negative: missing balances are properly handled (default value=0).
*big.Int can be nil but it expected not to be.
We're not going to do defensive nil check on every single reference value, else we'd double che size of the codebase (there are big ints everywhere)
row.Amount isn't nil by contract. Nullable fields are explicitly specified


> Bug 2 — getAssets (asset scaling) uses an over-broad prefix check
I thinkg this was a false negative (couldn't reproduce that) but I made the check more precise just to make sure

> Latent Bug 3 — getBalance queries with color but fetches from cache without it
Fair, fixed it


review again the new progress

NumaryBot (commented):

The patch introduces regressions in test spec initialization for scripts using metadata and in the MCP evaluate tool's advertised balances input shape.

internal/cmd/test_init.go
```diff
@@ -200,40 +201,52 @@ func runTestInitCmd(opts testInitArgs) error {
...
+			Amount:  new(big.Int).Set(amount),
+		})
+		stored[k] = struct{}{}
 	}
 
-	return outputMeta, nil
+	return balances, nil
+}
+
+func (s *TestInitStore) GetAccountsMetadata(c context.Context, q interpreter.MetadataQuery) (interpreter.AccountsMetadata, error) {
```
NumaryBot, line 250: 🟠 **[major] Preserve metadata stubs in test init**

When `numscript test init` runs a script that calls `meta(...)`, this delegated store now returns the empty `StaticStore.Meta` map instead of materializing the requested account/key pairs. That makes the interpreter raise `MetadataNotFound`, whereas the previous `TestInitStore` returned `""` for each queried metadata key so spec generation could complete. This should keep the test-init-specific stub behavior or populate `StaticStore.Meta` for the requested query.

internal/mcp_impl/handlers.go
NumaryBot, line 66: 🟠 **[major] Update MCP balances argument shape**

After `interpreter.Balances` became a list of balance rows, this binding expects `balances` to be a JSON array, but the MCP tool is still registered as an object and its description still tells clients to send the old nested map shape. MCP clients following the advertised schema (for example `{ "alice": { "USD/2": 100 } }`) will now fail in `BindArguments` instead of evaluating the script.

internal/specs_format/runner.go
```diff
@@ -155,6 +155,14 @@ func runRawSpec(stdout io.Writer, stderr io.Writer, rawSpec RawSpec) (Specs, Spe
 		return Specs{}, SpecsResult{}, false
 	}
 
+	// Reject a malformed specs file before running anything, just like the JSON
+	// parse error above.
+	if err := validateSpecs(specs); err != nil {
+		_, _ = stderr.Write([]byte(ansi.ColorRed(fmt.Sprintf("\nError: %s.specs.json\n\n", rawSpec.NumscriptPath))))
+		_, _ = stderr.Write([]byte(err.Error() + "\n"))
+		return Specs{}, SpecsResult{}, false
+	}
```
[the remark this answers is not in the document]
ascandone: it's ok, that's out of scope for now (we'll have to improve the return type from Check to handle errors that aren't about the assertions). Not a reproducible bug for the user or ledger

internal/interpreter/store.go
```diff
@@ -0,0 +1,82 @@
...
+	Balances Balances
+	Meta     AccountsMetadata
+}
+
+func (s StaticStore) GetBalances(_ context.Context, q BalanceQuery) (Balances, error) {
+	var output Balances
+	for _, item := range q {
+		baseAsset, isCatchAll := strings.CutSuffix(item.Asset, "/*")
+
+		if isCatchAll {
```
NumaryBot, line 40: 🟠 **[major] Treat empty wildcard balance results as zero balances**

When a scaling source preloads `ASSET/*` for an account that has no matching balances, this catch-all branch returns no rows at all. `InternalBalances` then has no entry for the queried account, so `SourceWithScaling` later fails the `acc, ok` check with `InvalidUnboundedAddressInScalingAddress` instead of treating the account as having zero funds and returning the normal missing-funds result. This affects `StaticStore` and any ledger store that returns an empty list for an empty wildcard query.

internal/interpreter/interpreter.go
```diff
@@ -431,19 +379,20 @@ func (st *programState) pushReceiver(name string, monetary *big.Int) {
 		postings := Posting{
 			Source:      sender.Name,
 			Destination: name,
-			Asset:       coloredAsset(st.CurrentAsset, &sender.Color),
+			Asset:       string(st.CurrentAsset),
```
NumaryBot, line 382: 🟠 **[major] Include posting color in movement aggregation**

Now that colored postings keep `Asset` uncolored and put the color in `Color`, specs using `expect.movements` will collapse transfers that have the same source, destination, and asset but different colors, because `getMovements` still aggregates only by `posting.Asset`. Previously `coloredAsset` made those movement keys distinct, so colored movement assertions can now silently pass or fail against the wrong aggregate.

internal/cmd/test_init.go
NumaryBot, line 250: 🟠 **[major] Preserve metadata stubbing in test-init**

For scripts that call `meta()`, `test init` constructs this store with an empty metadata map; delegating to `StaticStore` returns that empty map and ignores the requested keys, so `meta()` raises `MetadataNotFound`. The previous `TestInitStore` synthesized empty strings for every requested metadata key, allowing specs to be generated for scripts with metadata reads.

internal/mcp_impl/handlers.go
NumaryBot, line 66: 🟠 **[major] Advertise balances as an array to MCP clients**

This now decodes `balances` as `interpreter.Balances` (`[]BalanceRow`), but the MCP tool definition still declares `balances` with `WithObject` and the old nested-map example. MCP clients that follow the advertised schema will send an object that fails binding, while clients may be prevented from sending the new array shape.

NumaryBot (commented):

The patch introduces several functional regressions around empty wildcard balance queries, colored movement assertions, test initialization for metadata, and MCP balance argument binding. These affect existing workflows and should be fixed before considering the change correct.

internal/interpreter/balances.go
ascandone: done (that part was reworked)

internal/mcp_impl/handlers.go
ascandone: fixed

internal/cmd/test_init.go
NumaryBot, line 250: 🟠 **[major] Preserve metadata stubs in test-init**

When `numscript test-init` is run for scripts that call `meta(...)`, this now delegates to `StaticStore`, which returns the empty metadata map instead of synthesizing the requested account/key pairs as `TestInitStore` did before. In that scenario (for example `vars { string $s = meta(@a, "k") }`) `meta()` raises `MetadataNotFound`, so specs can no longer be generated for metadata-driven scripts.

internal/cmd/test_init.go
```diff
@@ -200,40 +201,52 @@ func runTestInitCmd(opts testInitArgs) error {
...
-func (s TestInitStore) GetBalances(_ context.Context, q interpreter.BalanceQuery) (interpreter.Balances, error) {
-	outputBalance := interpreter.Balances{}
-	for queriedAccount, queriedCurrencies := range q {
-
-		for _, curr := range queriedCurrencies {
-			amt := utils.NestedMapGetOrPutDefault(s.Balances, queriedAccount, curr, func() *big.Int {
-				return new(big.Int).Set(s.DefaultBalance)
-			})
+func (s *TestInitStore) GetBalances(ctx context.Context, q interpreter.BalanceQuery) (interpreter.Balances, error) {
+	balances, err := s.StaticStore.GetBalances(ctx, q)
```
NumaryBot, line 208: 🟠 **[major] Materialize wildcard balance queries in test-init**

For asset-scaling sources the interpreter queries a wildcard asset like `USD/*`; when the account has no stored rows, `StaticStore.GetBalances` returns an empty slice for that wildcard, so the loop below never seeds the account with `DefaultBalance`. In scripts using `source = @src with scaling through @swap`, `MakeSpecsFile` then reaches scaling with no cached balance entry and returns `InvalidUnboundedAddressInScalingAddress` instead of generating a spec.

internal/interpreter/interpreter.go
NumaryBot, line 31: 🟡 **[minor] Include posting colors in pretty output**

When a colored transfer is executed, this new `Color` field is populated but `PrettyPrintPostings` still renders only source, destination, asset, and amount. For `numscript run` with the default pretty output, two postings with the same asset and different colors now look identical; previously the color was embedded in the asset string, so the pretty output loses information unless JSON output is requested.

NumaryBot (commented):

The patch introduces regressions in `test-init` for metadata and asset-scaling scripts, causing spec generation to fail in supported scenarios. It also drops color information from the default pretty posting output.

internal/interpreter/asset_scaling.go
```diff
@@ -25,25 +25,12 @@ func buildScaledAsset(baseAsset string, scale int64) string {
...
-}
-
-func getAssets(balance AccountBalance, baseAsset string) map[int64]*big.Int {
+func getAssets(accountBalances []AccountBalance, baseAsset string) map[int64]*big.Int {
 	result := make(map[int64]*big.Int)
-	for asset, amount := range balance {
-		if strings.HasPrefix(asset, baseAsset) {
-			_, scale := getAssetScale(asset)
-			result[scale] = amount
+	for _, accBalance := range accountBalances {
```
NumaryBot, line 30: 🟠 **[major] Do not let scaling spend colored balances**

When a script has cached a colored balance for the same account/base asset and then uses `with scaling`, this loop includes that colored row as a conversion candidate even though the scaling path later emits and subtracts colorless postings. For example, a cached RED `USD/2` balance can be treated as available uncolored `USD/2`, allowing scaling to create postings the ledger cannot fund from uncolored balances.

internal/cmd/test_init.go
```diff
@@ -200,40 +201,52 @@ func runTestInitCmd(opts testInitArgs) error {
...
+	// (account, asset, color), so its output can't tell a known account from an
+	// unknown one. Track what we've actually funded ourselves instead.
+	stored := make(map[key]struct{}, len(s.StaticStore.Balances))
+	for _, b := range s.StaticStore.Balances {
+		stored[key{b.Account, b.Asset, b.Color}] = struct{}{}
 	}
 
-	return outputBalance, nil
-}
+	for i := range balances {
```
NumaryBot, line 223: 🟠 **[major] Seed defaults for catch-all balance queries**

For `test-init` on scripts using asset scaling, the batched query is a catch-all like `USD/*`; with an empty generated store, `StaticStore.GetBalances` returns no rows for that query, so this loop never adds the default balance. The interpreter then has no cached account entry and `MakeSpecsFile` fails with `InvalidUnboundedAddressInScalingAddress` instead of generating a spec.

NumaryBot (commented):

The patch introduces correctness regressions around colored balances interacting with asset scaling, and it breaks specs generation for scaling scripts with empty initial balances. These issues can produce invalid postings or prevent supported tooling from working.

internal/cmd/test_init.go
ascandone: true, but out of scope for this pr. It's a pre-existing bug, about an experimental, undocumented feature, and our pr has higher priority

internal/interpreter/interpreter.go
```diff
@@ -778,8 +733,9 @@ func (s *programState) tryTakingUpTo(source parser.Source, amount *big.Int) (*bi
 			return nil, err
 		}
 
-		baseAsset, assetScale := getAssetScale(s.CurrentAsset)
-		acc, ok := s.CachedBalances[*account]
+		baseAsset, assetScale := s.CurrentAsset.GetBaseAndScale()
+
+		acc, ok := s.CachedBalances[string(account)]
```
NumaryBot, line 738: 🟠 **[major] Handle empty wildcard balance results for scaling**

When a `with scaling` source queries a catch-all asset like `EUR/*` and the store returns no rows (valid when the source has no balances in that asset family), the cache never gets an entry for the account. This lookup then treats the account as invalid and returns `InvalidUnboundedAddressInScalingAddress` for an otherwise valid `source = @src with scaling through @swap` instead of using an empty asset set and producing the normal missing-funds/zero-send behavior; the same pattern exists in the send-all scaling path.

NumaryBot (commented):

The refactor leaves catch-all balance queries with no result rows unrepresented in the cache, which breaks scaling sources for empty asset families. The test-init store also no longer seeds defaults for those wildcard queries, so specs generation fails for scaling scripts.

internal/cmd/test_init.go
NumaryBot, line 223: 🟠 **[major] Seed defaults for wildcard test-init queries**

For `test-init` on scripts using asset scaling, the preload balance query is a catch-all such as `EUR/*`. If the generated store is empty, `StaticStore.GetBalances` returns an empty slice, so this loop never seeds `DefaultBalance` into `StaticStore`; `numscript test-init` then fails before generating specs for scaling scripts with no preexisting balances.

internal/interpreter/interpreter.go
ascandone: scaling is out of scope for the PR

internal/cmd/test_init.go
ascandone: scaling is out of scope for the PR

ascandone:
@coderabbitai review this. Forget about asset scaling: it's unstable and undocumented. Forget about pre-existing issues that are out of scope for the PR

internal/interpreter/interpreter.go
NumaryBot, line 738: 🟠 **[major] Treat empty scaling balance queries as empty balances**

When a scaling source queries `ASSET/*` and the store returns no rows for a valid non-`world` account, the row-based cache has no way to record that the account was queried, so `ok` is false here. This makes scripts like `send [USD 1] (source = @src with scaling through @swap ...)` fail with `InvalidUnboundedAddressInScalingAddress` whenever `src` simply has no matching balances, instead of continuing with an empty balance set and producing the normal missing-funds behavior; the same lookup pattern is used in the send-all scaling branch above.

NumaryBot (commented):

The row-based balance cache regresses scaling sources when wildcard balance queries return no rows, causing valid zero-balance accounts to be reported as invalid. This is a functional behavior issue in a supported feature path.

internal/interpreter/balances.go
```diff
@@ -2,110 +2,98 @@ package interpreter
...
-	for account, accountBalances := range b {
-		for asset, amount := range accountBalances {
-			utils.NestedMapGetOrPutDefault(cloned, account, asset, func() *big.Int {
-				return new(big.Int).Set(amount)
-			})
-		}
-	}
-	return cloned
+type BalanceRow struct {
+	Account string   `json:"account"`
```
gfyrag, line 11: Introducing a BalanceIdentity aggregating Asset, Account, Color and any future property identifying a balance would simplify some comparison over this pr, while preventing you to have to modify each comparison the next time you add a property.

internal/interpreter/balances.go
```diff
@@ -2,110 +2,98 @@ package interpreter
...
+type Balances []BalanceRow
+
+// FirstDuplicate returns the first row whose (account, asset, color) key already
+// appeared earlier in the list, if any. That triple is the identity of a balance
+// entry and the amount is its value, so a repeated key is an ambiguous,
+// malformed input.
+func (rows Balances) FirstDuplicate() (BalanceRow, bool) {
+	seen := make(map[[3]string]struct{}, len(rows))
+	for _, row := range rows {
+		key := [3]string{row.Account, row.Asset, row.Color}
```
gfyrag, line 25: Could use BalanceIdentity

internal/interpreter/balances.go
```diff
@@ -2,110 +2,98 @@ package interpreter
...
-		cachedAcc := utils.MapGetOrPutDefault(b, acc, func() AccountBalance {
-			return AccountBalance{}
-		})
-
-		for curr, amt := range accBalances {
-			cachedAcc[curr] = amt
+// findRow returns the amount for a given (account, asset, color), if present.
+func findRow(rows Balances, account, asset, color string) (*big.Int, bool) {
+	for i := range rows {
+		if rows[i].Account == account && rows[i].Asset == asset && rows[i].Color == color {
```
gfyrag, line 65: Could use BalanceIdentity

internal/interpreter/batch_balances_query.go
```diff
@@ -50,15 +50,19 @@ func (st *programState) findBalancesQueriesInStatement(statement parser.Statemen
 	}
 }
 
-func (st *programState) batchQuery(account string, asset string, color *string) {
+func (st *programState) batchQuery(account AccountAddress, asset Asset, color String) {
```
gfyrag, line 53: Could use BalanceIdentity

internal/interpreter/function_exprs.go
```diff
@@ -83,37 +84,38 @@ func balance(
 	s *programState,
 	r parser.Range,
 	args []Value,
-) (*Monetary, InterpreterError) {
+) (Monetary, InterpreterError) {
```
gfyrag, line 87: Don't really understand why this type of change.
It complicates the review, while not being a very good practice in go.
Not a big deal.

internal/interpreter/internal_balances.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return false
+}
+
+// given a BalanceQuery, return a new query which only contains needed
+// (account, asset, color) tuples (that is, the ones that aren't already cached)
+func (b InternalBalances) filterQuery(q BalanceQuery) BalanceQuery {
+	filteredQuery := BalanceQuery{}
+	for _, item := range q {
+		if !b.has(item.Account, item.Asset, item.Color) {
```
gfyrag, line 99: Probably again a good opportunity for a BalanceIdentity.

gfyrag (dismissed):
Feels ok.
There is multiple times when you need to split assets over "/".
I start to think the ledger and numscript could not have to do this kind of operation.
We could add a "precision" property.
A range could be a good thing at BalanceQuery level too as I think you have needs for that.

The final USD/2, is just a display detail, internally, we can have ASSET = USD, PRECISION = 2

I think this PR put the fundation for that.

internal/interpreter/function_exprs.go
ascandone: It simplifies the previous code, which used a ptr in order to be able to return an absent value as well when we have an error, with the polymorphic combinators as well (the ones operating on [T]).
But it was a naive choice done at the time, because we can actually return a default value even with generics:
```go
var default_ T
return default_
```
This change itself isn't related with the PR's goal but this part of the code was involved so I simplified that as well

ascandone: I think I'm missing the reason why this isn't a good practice, but if the reason is unnecessary allocs, it only happens in case of errors, which means at most once per exec

internal/interpreter/interpreter.go
NumaryBot, line 738: 🟠 **[major] Handle empty scaled-source queries as empty balances**

When a scaled source has no matching uncolored balances, the new row-based balance query returns no `BalanceRow`, so `CachedBalances` has no entry for that account. This makes a normal account such as `@src with scaling through @swap` fail with `InvalidUnboundedAddressInScalingAddress` instead of being treated as an empty balance and producing missing funds/no postings; the same check exists in the send-all scaling branch. This error should only apply to `@world`, not to accounts whose catch-all query returned an empty list.

NumaryBot (commented):

The row-based balance representation loses the ability to distinguish an empty catch-all balance result from an invalid `@world` scaling source, causing valid scaled transfers from empty accounts to fail with the wrong interpreter error.

internal/interpreter/interpreter.go
NumaryBot, line 31: 🟠 **[major] Preserve color when computing movement assertions**

When a spec uses `expect.movements` for colored transfers, moving color out of `Asset` into this new `Color` field means `getMovements` still aggregates only by source/destination/asset and collapses distinct colors together. For example, `COIN` RED 10 and `COIN` BLUE 5 between the same accounts are reported as a single `COIN` movement of 15, so colored movement assertions can pass or fail incorrectly.

NumaryBot (commented):

The patch adds first-class posting colors but leaves movement assertions aggregating without color, which breaks correctness for specs that assert movements involving colored postings.

NumaryBot (commented):

The patch mishandles empty catch-all balance queries used by asset scaling, causing valid zero-balance accounts to be classified as invalid and breaking initialization for those scripts.

internal/interpreter/store.go
NumaryBot, line 40: 🟠 **[major] Materialize empty catch-all balance queries**

When a scaling source queries an account with no matching balances (for example `@src with scaling through @swap` and no `EUR`/`EUR/*` rows), this catch-all branch appends nothing. The cache then has no entry for `src`, so the interpreter reports `InvalidUnboundedAddressInScalingAddress` instead of treating the account as having zero balance and returning the usual missing-funds path; `test-init` also cannot seed a default balance because it only iterates over rows returned here.

NumaryBot (commented):

The new row-based balance store drops empty wildcard query results, which changes scaling-source error behavior and breaks default balance generation for scaling cases.

internal/interpreter/store.go
NumaryBot, line 40: 🟠 **[major] Preserve empty wildcard balance queries**

When a wildcard balance query such as `USD/*` matches no uncolored rows, this branch returns no row at all, so `CachedBalances` never gets an entry for the queried account. A scaling source from an account with no matching uncolored balances then fails with `InvalidUnboundedAddressInScalingAddress` instead of treating the account as having zero available funds and producing the expected missing-funds path; this also prevents `test init` from default-funding scaling sources.

ascandone: out of scope. Won't fix

gfyrag (approved):

PR #163, opened 2026-06-19
feat: script builder

ascandone:


builder/expression.go
```diff
@@ -0,0 +1,60 @@
+package builder
+
+import "math/big"
+
+type Expression[T ExprType] render
+
+func ExprVar[T ExprType](v *Var[T]) Expression[T] {
+	return func(env *env, w int) {
+		if !v.set {
```
NumaryBot: 🟠 **[major] Allocate external vars per program build**

When the same `Var` or prebuilt `Statement` is passed to `BuildProgram` more than once, this cached `set/name` state is reused with a fresh `env`, so the new pools are not advanced and the generated script can reference `$account_0` without declaring it, or collide with the first literal allocated in that build. This makes rendering non-idempotent for reusable builder values.

builder/vars.go
```diff
@@ -0,0 +1,63 @@
...
+func NewNumberVar() Var[ExprTypeNumber] {
+	return Var[ExprTypeNumber]{
+		alloc: func(env *env) string {
+			id := env.numbersPool.getFreshId()
+			return numberToName(id)
+		},
+	}
+}
+
+func (v Var[ExprTypeAccount]) FillAccount(account string) (string, string) {
```
NumaryBot: 🟠 **[major] Preserve type safety for Fill methods**

These receivers do not specialize the method to `Var[ExprTypeAccount]`; in Go they declare a fresh receiver type parameter, so every `Var[T]` gets every `Fill*` method. A caller can therefore compile `NewAssetVar().FillAccount(...)`, producing an account value for an `asset` variable and failing later at runtime instead of being rejected by the builder API.

NumaryBot (commented):

The new builder can emit invalid scripts when builder values are reused, and its typed variable API does not enforce the advertised variable type at compile time. These are correctness issues in the newly introduced public API.

NumaryBot (commented):
## ✅ Approve — automated review

No discrete correctness issues were identified in the introduced builder implementation based on the diff and repository context.

_No findings._

NumaryBot (commented):
## ✅ Approve — automated review

No discrete correctness issues were identified in the added builder package based on the diff and surrounding parser/runtime expectations.

_No findings._

NumaryBot (commented):
## ✅ Approve — automated review

No discrete correctness issues were found in the added builder implementation based on the diff and existing Numscript grammar/API expectations.

_No findings._

builder/source.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+// SrcAllowingUnboundedOverdraft wraps a source (typically SrcAccount or
+// SrcColored) and appends the `allowing unbounded overdraft` clause, per the
+// Numscript grammar rule `srcAccountUnboundedOverdraft`
+// (address colorConstraint? ALLOWING UNBOUNDED OVERDRAFT). It lets a non-world
+// source go negative — required, for instance, when minting a colour the
+// account does not yet hold.
+func SrcAllowingUnboundedOverdraft(source Source) Source {
```
NumaryBot: 🟠 **[major] Restrict overdraft wrapper to account sources**

When callers wrap a composite source such as `SrcInorder(...)`, this appends `allowing unbounded overdraft` after a `{ ... }` source, producing syntax that does not match the Numscript grammar because the clause is only valid after `address colorConstraint?`. Since the typed builder API accepts any `Source` here, valid-looking builder calls can generate unparsable scripts.

NumaryBot (commented):

The new builder can generate invalid Numscript for a typed API call by applying the unbounded overdraft clause to source forms where the grammar does not allow it.

NumaryBot (commented):
## ✅ Approve — automated review

I did not identify any discrete correctness issues in the added builder API from the diff. The generated scripts and variable binding approach appear consistent with the existing Numscript grammar and runtime expectations.

_No findings._

Azorlogh (dismissed):

Azorlogh (approved):

NumaryBot (commented):

The builder can generate colored-source scripts that are rejected by the default interpreter execution path because the required feature flag is not included or surfaced.

builder/source.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func SrcColored(
+	accountExpr Expression[ExprTypeAccount],
+	colorExpr Expression[ExprTypeString],
+) Source {
+	return func(env *env, w int) {
+		accountExpr(env, w)
+		env.builder.WriteString(" \\ ")
```
NumaryBot, line 40: 🟠 **[major] Enable asset-colors when rendering colored sources**

When callers use `SrcColored`/`SrcColoredOverdraft` and execute the returned script via the normal `Parse(...).Run` path, the interpreter rejects it because colored sources require the `experimental-asset-colors` feature flag, but `BuildProgram` never emits a `#![feature(...)]` declaration or returns required flags. The builder should track this feature or otherwise expose a way to make generated colored scripts runnable.

PR #164, opened 2026-06-23
feat(builder): add SrcAllowingUnboundedOverdraft combinator

paul-nicolas:
Adds the `allowing unbounded overdraft` source combinator on top of #163, required by transaction-plane EN-1271. Base is #163's branch.

### DIFF
--- a/builder/source.go
+++ b/builder/source.go
@@ -17,6 +17,19 @@ func SrcColored(
 	}
 }
 
+// SrcAllowingUnboundedOverdraft wraps a source (typically SrcAccount or
+// SrcColored) and appends the `allowing unbounded overdraft` clause, per the
+// Numscript grammar rule `srcAccountUnboundedOverdraft`
+// (address colorConstraint? ALLOWING UNBOUNDED OVERDRAFT). It lets a non-world
+// source go negative — required, for instance, when minting a colour the
+// account does not yet hold.
+func SrcAllowingUnboundedOverdraft(source Source) Source {
+	return func(env *env, w int) {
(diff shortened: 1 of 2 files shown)


builder/source.go
```diff
@@ -17,6 +17,19 @@ func SrcColored(
 	}
 }
 
+// SrcAllowingUnboundedOverdraft wraps a source (typically SrcAccount or
+// SrcColored) and appends the `allowing unbounded overdraft` clause, per the
+// Numscript grammar rule `srcAccountUnboundedOverdraft`
+// (address colorConstraint? ALLOWING UNBOUNDED OVERDRAFT). It lets a non-world
+// source go negative — required, for instance, when minting a colour the
+// account does not yet hold.
+func SrcAllowingUnboundedOverdraft(source Source) Source {
```
NumaryBot, line 26: 🟠 **[major] Restrict overdraft combinator to account sources**

When callers wrap any composite source such as `SrcInorder(...)` or `SrcCapped(...)`, this exported API still appends `allowing unbounded overdraft`, producing scripts like `{ ... } allowing unbounded overdraft` or `max ... from ... allowing unbounded overdraft` that the Numscript grammar cannot parse. The clause is only valid on the account/color source form, so the builder should avoid accepting arbitrary `Source` values here or otherwise prevent invalid compositions.

NumaryBot (commented):

The new exported combinator can generate unparsable Numscript for valid Go calls that pass non-account sources. This should be corrected before considering the patch fully correct.

PR #169, opened 2026-07-08
feat: resolve dependencies

ascandone:
This PR:
1. (refactor) extracts an utility that evaluates expressions and handles vars on its own (`internal/interpreter/evaluate_expr.go`), so that the interpreter can eval exprs using that one instead of the whole interpreter state 
2. (feat) implements a static analysis that (after vars substitutions, given some balances) finds what accounts balances we need to read/write, what account meta we need to read/write, and what tx meta we need to write

The exact algo for balances tracking is:
1. we register account destinations as writes. We mark each dest with each color that appears anywhere in the source
2. we register account sources as writes+reads if they have bounded overdraft, or as writes if they are unbounded overdraft.
    1. note: if `allowing unbounded overdraft` or `allowing overdraft up to ..` isn't specified, it counts as bounded overdraft, unless it's `@world`. Otherwise, that's the source of truth. For example: the `@world allowing overdraft up to [ASSET 0]` counts as bounded overdraft (recorded as read+write)
3. we register balance() and overdraft() calls as (uncolored) reads

In order to get 3. properly, we make sure we eval every single expr of the ast (even though it's not used as source or dest). For example, a "max" clause could be using a balance() expr, which we need to track

Note that we explicitly avoid scaling to be implemented, and output an error. That's ok, as scaling isn't documented yet, nor properly specified


internal/interpreter/resolve_dependencies.go
```diff
@@ -0,0 +1,341 @@
...
+		p := NewArgsParser(args)
+		key := parseArg(p, fnCall.Range, expectString)
+		_ = parseArg(p, fnCall.Range, expectAnything)
+		if err := p.parse(); err != nil {
+			return err
+		}
+		st.deps.TxMetaWrites[string(key)] = struct{}{}
+		return nil
+
+	default:
```
NumaryBot, line 183: 🟠 **[major] Unsupported top-level function calls silently succeed in ResolveDependencies**

When a script contains a top-level function call other than `set_account_meta` or `set_tx_meta`, `ResolveDependencies` falls through to a default branch that evaluates arguments and returns success. The interpreter would reject the same statement with `UnboundFunctionErr`, so callers using dependency resolution as a preflight check will incorrectly proceed with resolved dependencies for a script that cannot actually run (e.g. `foo(balance(@a, USD))` resolves without error).

**Suggestion:** Add an explicit error return in the default/fallback branch for unrecognised function names, mirroring the `UnboundFunctionErr` that the interpreter itself raises. This ensures `ResolveDependencies` and the interpreter agree on which scripts are valid.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/169#issuecomment-4927654887

internal/interpreter/resolve_dependencies.go
NumaryBot, line 183: 🟠 **[major] Unrecognised top-level function calls silently succeed in ResolveDependencies**

When a script contains a top-level function call other than `set_account_meta` or `set_tx_meta`, the default branch in `ResolveDependencies` evaluates the arguments and returns success. The interpreter, however, rejects the same statement with `UnboundFunctionErr`. This inconsistency means callers using `ResolveDependencies` as a preflight validation step will incorrectly obtain resolved dependencies for a script that will fail at runtime.

**Suggestion:** Return an explicit `UnboundFunctionErr` (or equivalent) from the default/fallback branch for unrecognised function names, mirroring the interpreter's own behaviour and keeping both code paths consistent.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/169#issuecomment-4927654887

internal/interpreter/function_exprs.go
```diff
@@ -3,16 +3,67 @@ package interpreter
...
+	args, err := evaluateExpressions(env, fnCall.Args)
+	if err != nil {
+		return nil, err
+	}
+
+	switch fnCall.Caller.Name {
+	case analysis.FnVarOriginMeta:
+		if type_ == nil {
+			return nil, InvalidNestedMeta{}
+		}
```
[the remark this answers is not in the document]
ascandone: neh

NumaryBot (commented):

Resolved 1 stale NumaryBot review thread (1 fixed, 0 outdated).

Summary: https://github.com/formancehq/numscript/pull/169#issuecomment-4927654887

Azorlogh:
This seems to miss a metadata read
```go
func TestResolveMultipleMetaKeysAreAllReads(t *testing.T) {
      deps := resolve(t, `
              vars {
                      account $x = meta(@config, "recipient")
                      account $y = meta(@config, "fallback")
              }
              send [USD 10] (
                      source = $x allowing unbounded overdraft
                      destination = $y
              )
      `, nil, interpreter.StaticStore{
              Meta: interpreter.AccountsMetadata{
                      {Account: "config", Key: "recipient", Value: "alice"},
                      {Account: "config", Key: "fallback", Value: "bob"},
              },
      })

      require.Equal(t, map[interpreter.MetaDependency]struct{}{
              {Account: "config", Key: "recipient"}: {},
              {Account: "config", Key: "fallback"}:  {},
      }, deps.MetaReads)
}
```

Azorlogh (approved):

PR #171, opened 2026-07-10
fix: properly version json schemas

ascandone:
On the previous PRs we refactored the specs and inputs schema by doing breaking changes. The change isn't realised yet so it still doesn't break user codebases (and we are still in time to publish utilities to update the schema) but it will break editor diagnostics that use the $schema field.

Fixed it by adding a v1 prefix in the new format, so that we can leave the old format in the repo and user will be able to keep referring to that

---

Pr can be read commit by commit

### DIFF
--- a/inputs.schema.json
+++ b/inputs.schema.json
@@ -21,35 +21,18 @@
   },
   "definitions": {
     "Balances": {
-      "type": "array",
-      "description": "List of account balances. The (account, asset, color, scope) tuple of each entry must be unique within the list.",
-      "items": { "$ref": "#/definitions/BalanceRow" }
-    },
-
-    "BalanceRow": {
       "type": "object",
-      "description": "The balance of a given (account, asset, color, scope)",
+      "description": "Map of account names to asset balances",
       "additionalProperties": false,
-      "required": ["account", "asset", "amount"],
-      "properties": {
-        "account": {
-          "type": "string",
-          "pattern": "^([a-zA-Z0-9_-]+(:[a-zA-Z0-9_-]+)*)$"
-        },
-        "asset": {
-          "type": "string",
-          "pattern": "^([A-Z]+(/[0-9]+)?)$"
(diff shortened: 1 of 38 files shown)


inputs.schema.json
```diff
@@ -21,35 +21,18 @@
   },
   "definitions": {
     "Balances": {
```
NumaryBot, line 23: 🟠 **[major] Keep the inputs schema aligned with CLI JSON**

When users follow `inputs.schema.json`, this now validates `balances` as an account-to-asset object map, but `internal/cmd/run.go` still unmarshals `balances` into `interpreter.Balances` (`[]BalanceRow`) and `metadata` into `AccountsMetadata` (`[]AccountMetadataRow`). A file accepted by this schema will therefore fail at runtime with a JSON unmarshal error, while the currently supported array form is marked invalid by the unversioned schema.

specs.schema.json
```diff
@@ -83,35 +83,18 @@
     },
 
     "Balances": {
```
NumaryBot, line 85: 🟠 **[major] Keep the specs schema aligned with test specs**

This changes the default specs schema's `Balances` definition to an object map, but the specs runner still decodes `balances`, `expect.endBalances`, and related fields as `interpreter.Balances` slices. Specs that validate against `specs.schema.json` will fail to load in the current runner, and existing valid specs that do not opt into `v1.specs.schema.json` will be reported as invalid by the default schema.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/171#issuecomment-4933557841

flemzord (approved):

inputs.schema.json
ascandone: false negative

specs.schema.json
ascandone: false negative

PR #172, opened 2026-07-14
vm verifier

ascandone:


internal/vm/verify.go
```diff
@@ -0,0 +1,359 @@
...
+		readOpt(bankInt, ext.A)
+		readOpt(bankStr, ext.B)
+		d.reads = append(d.reads, currentAssetRef)
+		write(bankInt, instr.A)
+	case Op_SendToAccount:
+		readOpt(bankStr, instr.A)
+		readOpt(bankInt, instr.B)
+		readOpt(bankStr, instr.C)
+		d.reads = append(d.reads, currentAssetRef)
+	case Op_MkAllotment:
```
NumaryBot, line 249: 🔴 **[blocker] Zero-length Op_MkAllotment passes verification but can panic at execution**

When Op_MkAllotment is encoded with C == 0 (zero-length allotment), the verification loops record no accesses to the start indices in instr.A or instr.B. The instruction passes verification even when only the amount register has been allocated. At execution time, Exec evaluates slices such as vm.intsRegs[255:255] and vm.portionsRegs[255:255] against the dynamically-sized register banks, which are shorter, causing a panic. This breaks the safety guarantee that verified bytecode cannot crash the VM.

**Suggestion:** Either explicitly reject zero-length allotments in the verifier (return an error if C == 0), or ensure the verifier accounts for the start indices instr.A and instr.B even when C == 0 so that out-of-range accesses are caught before execution.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/172#issuecomment-4970923990

ascandone:
closing: already integrated into the main pr

PR #178, opened 2026-08-04
fix(mcp_impl): reject balance amounts beyond the safe JSON integer range

ascandone:
Supersedes #142 and #143 (closing those, see comments there for why they no longer apply as-is).

## Background

#142/#143 pinned/fixed a float64-truncation bug in `parseBalancesJson`, a manual JSON parser for the `evaluate` MCP tool's balances that no longer exists — it was removed when the tool switched to `request.BindArguments` into a typed `interpreter.Balances` (with `Amount *big.Int`).

That switch does fix the fractional-amount case: `math/big.Int`'s own `UnmarshalJSON` rejects `"100.9"` outright.

It does **not** fix the large-magnitude case, though — it just moved it. I confirmed by reproducing the actual request path: the MCP server decodes the incoming JSON-RPC message via `json.Unmarshal(message, &request)` into `Params.Arguments any`, and Go's default for JSON numbers into `any` is always `float64`, regardless of magnitude. That happens *before* `BindArguments` ever runs. So an amount like `9007199254740993` (one past float64's exact range, 2^53-1) is silently rounded to `9007199254740992` at the transport layer, and by the time `BindArguments` sees it, it looks like a perfectly clean, valid integer — `big.Int` has no way to tell it was ever anything else.

## Fix

Added `checkBalanceAmountsInSafeRange`, which inspects the raw (pre-`BindArguments`) float64 amounts via `request.GetArguments()` and rejects anything outside `±(2^53-1)`, before the value is converted to `*big.Int` and the corruption becomes unobservable. Mirrors the bounds-check approach from the old #143, just applied at the current call site.

## Test plan

- [x] `go test ./internal/mcp_impl/...` — all existing tests pass, plus new coverage:
  - rejects an amount one past the safe range, reproduced via an actual JSON-RPC message round-tripped through `json.Unmarshal` (not just a hand-built Go struct) to match the real corruption path
  - rejects a large negative amount
  - accepts amounts within the safe range
  - still rejects fractional amounts (regression check that this path is unaffected)

### DIFF
--- a/internal/mcp_impl/handlers.go
+++ b/internal/mcp_impl/handlers.go
@@ -25,6 +25,7 @@ func addEvalTool(s *server.MCPServer) {
 		mcp.WithArray("balances",
 			mcp.Required(),
 			mcp.Description(`The accounts' balances. A list of entries, each an object with an "account", an "asset", an integer "amount", an optional "color", and an optional "scope".
+			"amount" must be an integer with magnitude at most 2^50-1 (1125899906842623).
 			The (account, asset, color, scope) tuple of each entry must be unique within the list.
 			For example: [ { "account": "alice", "asset": "USD/2", "amount": 100 }, { "account": "alice", "asset": "EUR/2", "amount": -42 }, { "account": "bob", "asset": "BTC", "amount": 1 } ]
 			`),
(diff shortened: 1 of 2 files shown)


internal/mcp_impl/handlers.go
```diff
@@ -48,6 +48,48 @@ func addEvalTool(s *server.MCPServer) {
...
+		return nil
+	}
+
+	for _, rowRaw := range balancesRaw {
+		row, ok := rowRaw.(map[string]any)
+		if !ok {
+			continue
+		}
+
+		amount, ok := row["amount"].(float64)
```
NumaryBot: 🟠 **[major] Reject rounded fractional amounts near the safe boundary**

For JSON numbers just over the safe bound that round back inside it, such as `9007199254740991.1` decoding to the float64 value `9007199254740991`, this comparison passes and `BindArguments` then accepts the rounded integer as a `big.Int`. Because the original token is already lost here, this still allows silently corrupted out-of-range/fractional balances; the check needs access to the raw JSON/json.Number value or another conservative rejection strategy before binding.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/178#issuecomment-5179378939

Azorlogh (dismissed):

internal/mcp_impl/handlers.go
```diff
@@ -48,6 +50,21 @@ func addEvalTool(s *server.MCPServer) {
...
+// maxBalanceAmount bounds the magnitude of balance amounts. The MCP
+// transport decodes incoming JSON numbers into a generic float64 (via
+// Params.Arguments any) before BindArguments converts one to *big.Int, so an
+// amount too large - or with enough fractional digits - can silently round
+// to a different, smaller-looking whole number first. float64 represents
+// every integer up to 2^53-1 exactly, but a fractional amount can still
+// round away to a whole number anywhere its magnitude is large relative to
+// its fractional part, not just past 2^53-1: e.g. at 2^50 or above, a single
+// stray decimal digit (like "9007199254740991.1") already rounds cleanly to
+// a whole number. Below 2^50, float64's absolute precision is finer than 1,
```
NumaryBot: 🔴 **[blocker] Reject rounded fractional amounts near the safe boundary**

Amounts below this cap can still lose a fractional component during the initial `float64` decode. For example, `1125899906842623.01` rounds to the in-range integer `1125899906842623`, so `BindArguments` and this magnitude check both accept corrupted input. Validation must inspect the raw JSON number rather than relying on a reduced magnitude cap.

internal/mcp_impl/handlers.go
```diff
@@ -72,6 +89,15 @@ func handleEvalTool(ctx context.Context, request mcp.CallToolRequest) (*mcp.Call
 		return mcp.NewToolResultError(err.Error()), nil
 	}
 
+	for _, row := range args.Balances {
```
NumaryBot: 🔴 **[blocker] Handle missing balance amounts before comparing**

When a balance entry omits `amount` or supplies `null`, `BindArguments` leaves `row.Amount` nil and this call panics. Such malformed client input should produce a tool error rather than crashing the handler; check for nil before calling `CmpAbs`.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/178#issuecomment-5179378939

NumaryBot (commented):

Resolved 2 stale NumaryBot review threads (0 fixed, 2 outdated).

Summary: https://github.com/formancehq/numscript/pull/178#issuecomment-5179378939

Azorlogh (approved):

PR #181, opened 2026-08-06
feat: scoped accounts

ascandone:


NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/181#issuecomment-5207109296

internal/ir/assemble.go
```diff
@@ -440,7 +442,18 @@ func (i Save) assemble(a *assembler) error {
 	if err != nil {
 		return err
 	}
+	scope, err := a.optionalReg((*assembler).strReg, i.Scope)
+	if err != nil {
+		return err
+	}
+
 	a.emit(vm.Op_Save, account, asset, amount)
```
NumaryBot, line 450: 🔴 **[blocker] Bump the bytecode format version for changed opcode layouts**

Adding mandatory extension words to existing opcodes such as `Save`, metadata reads, and `Balance` changes the v2 instruction layout, but `FormatVersion` remains 2. Consequently, the new decoder accepts old v2 bytecode and the VM consumes the following real instruction as an extension word (or reads past the end), while older runtimes also accept newly emitted v2 files but misinterpret their extensions. Increment the format version so incompatible bytecode is rejected.

PR #182, opened 2026-08-06
feat: expose migration tool

ascandone:
## `numscript test --migrate`

Specs files written against older numscript versions no longer parse — `test` just fails with a JSON error. This adds a flag that rewrites them for you, then runs the tests on the updated content in the same pass.

```console
$ numscript test ./examples
Error: examples/main.num.specs.json
json: cannot unmarshal object into Go struct field Specs.balances of type interpreter.Balances

$ numscript test --migrate ./examples
↻ migrated examples/main.num.specs.json
✓ examples/main.num (1 tests)
```

Opt-in: without the flag, `test` never touches your files.

It fixes two things:

**The `v0.0.24` layout**, where balances and metadata were objects keyed by account, into today's row arrays:

```json
// before
"balances": { "world": { "USD/2": 0 } }

// after
"balances": [{ "account": "world", "asset": "USD/2", "amount": 0 }]
```

**A missing or outdated `$schema`**, pointed at the current one.

Files that are already current are left byte-for-byte untouched. Files that aren't get fully reformatted to match what `test-init` generates.

Known gap: `v0.0.24` encoded asset colors into the name (`USD_RED`). That can't be told apart from an asset name containing an underscore, so it's carried through unchanged and needs a manual fix.

## Metadata values are plain strings again

Reverts the unreleased tagged-value format. Values are now written as you'd write them in source:

```json
// before
{ "key": "portion", "value": { "type": "portion", "numerator": "3", "denominator": "25" } }

// after
{ "key": "portion", "value": "3/25" }
```

Metadata is untyped, so `set_tx_meta("k", "42")` and `set_tx_meta("k", 42)` are the same thing. Since the tagged format was never released, the migration tool targets `v0.0.24` instead.

## Also

- Schema documents `expect.error.negativeAmount` (already supported in code).
- The repo's own `testdata` was regenerated with the tool — that's most of the diff.
- Tests: unit tests for the migration cases, plus e2e tests that build and exec the real binary.

### DIFF
--- a/internal/specs_format/migrate.go
+++ b/internal/specs_format/migrate.go
@@ -1,0 +1,39 @@
+package specs_format
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+// MigrateSpecsContent rewrites a specs file's raw JSON to the current format:
+// a missing or non-canonical $schema is pointed at SchemaURL. It reports
+// whether anything changed, so a caller can gate writing the result behind a
+// flag while still surfacing that a file is stale.
+//
+// Staleness is detected structurally (just the $schema field), not by
+// comparing byte-for-byte against a canonical re-marshal — otherwise a file
+// that is already current, just formatted differently (e.g. one row per
+// line), would be falsely flagged. Once a file IS stale, though, the rewrite
+// re-marshals the whole struct, so it also canonicalizes formatting to match
+// what the rest of the CLI generates (e.g. test-init) — this is not a
+// minimal-diff patch.
+func MigrateSpecsContent(raw []byte) (out []byte, changed bool, err error) {
+	var specs Specs
+	if err := json.Unmarshal(raw, &specs); err != nil {
+		return nil, false, err
+	}
+
+	if strings.HasSuffix(specs.Schema, "/v1.specs.schema.json") {
+		return raw, false, nil
+	}
+
+	specs.Schema = SchemaURL
+
+	marshaled, err := json.MarshalIndent(specs, "", "  ")
+	if err != nil {
+		return nil, false, err
+	}
+	marshaled = append(marshaled, '\n')
+
+	return marshaled, true, nil
+}
--- a/internal/specs_format/migrate.go
+++ b/internal/specs_format/migrate.go
@@ -6,24 +6,26 @@ import (
 )
 
 // MigrateSpecsContent rewrites a specs file's raw JSON to the current format:
-// a missing or non-canonical $schema is pointed at SchemaURL. It reports
-// whether anything changed, so a caller can gate writing the result behind a
-// flag while still surfacing that a file is stale.
+// v0.0.24's nested-map balances/metadata are flattened to today's row arrays
+// (see migrate_v0024.go), and a missing or non-canonical $schema is pointed
+// at SchemaURL. It reports whether anything changed, so a caller can gate
+// writing the result behind a flag while still surfacing that a file is
+// stale.
 //
-// Staleness is detected structurally (just the $schema field), not by
-// comparing byte-for-byte against a canonical re-marshal — otherwise a file
-// that is already current, just formatted differently (e.g. one row per
-// line), would be falsely flagged. Once a file IS stale, though, the rewrite
-// re-marshals the whole struct, so it also canonicalizes formatting to match
-// what the rest of the CLI generates (e.g. test-init) — this is not a
-// minimal-diff patch.
+// $schema staleness is detected structurally, not by comparing byte-for-byte
+// against a canonical re-marshal — otherwise a file that is already current,
+// just formatted differently (e.g. one row per line), would be falsely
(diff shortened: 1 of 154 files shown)


NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/182#issuecomment-5425452721

internal/specs_format/migrate.go
```diff
@@ -0,0 +1,39 @@
...
+// Staleness is detected structurally (just the $schema field), not by
+// comparing byte-for-byte against a canonical re-marshal — otherwise a file
+// that is already current, just formatted differently (e.g. one row per
+// line), would be falsely flagged. Once a file IS stale, though, the rewrite
+// re-marshals the whole struct, so it also canonicalizes formatting to match
+// what the rest of the CLI generates (e.g. test-init) — this is not a
+// minimal-diff patch.
+func MigrateSpecsContent(raw []byte) (out []byte, changed bool, err error) {
+	var specs Specs
+	if err := json.Unmarshal(raw, &specs); err != nil {
```
NumaryBot: 🔴 **[blocker] Migrate tagged metadata before decoding the current format**

For specs created by the previous release, metadata values are tagged JSON objects, but unmarshalling directly into the new string-valued `Specs` fails before any migration occurs. Files already referencing `v1.specs.schema.json` are also returned unchanged, so `test --migrate` cannot migrate the exact prior format this change replaces; decode the legacy representation and convert its metadata values before testing whether migration is needed.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/182#issuecomment-5425452721

internal/specs_format/migrate.go
```diff
@@ -0,0 +1,61 @@
...
+// Once a file IS stale, the rewrite re-marshals the whole struct, so it also
+// canonicalizes formatting to match what the rest of the CLI generates (e.g.
+// test-init) — this is not a minimal-diff patch.
+func MigrateSpecsContent(raw []byte) (out []byte, changed bool, err error) {
+	specs, structureChanged, err := parseSpecsForMigration(raw)
+	if err != nil {
+		return nil, false, err
+	}
+
+	if !structureChanged {
```
NumaryBot: 🟠 **[major] Update stale schemas even when structure is current**

When a specs file has the current row layout but a missing or outdated `$schema`, this early return reports no change and leaves the file untouched. Thus `numscript test --migrate` does not perform the advertised schema migration; include `specs.Schema != SchemaURL` in the change decision before returning the original bytes.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/182#issuecomment-5425452721

internal/specs_format/migrate.go
```diff
@@ -0,0 +1,71 @@
...
+// test-init) — this is not a minimal-diff patch.
+func MigrateSpecsContent(raw []byte) (out []byte, changed bool, err error) {
+	specs, structureChanged, err := parseSpecsForMigration(raw)
+	if err != nil {
+		return nil, false, err
+	}
+
+	colorsDecoded := decodeLegacyColors(&specs)
+
+	if !structureChanged && !colorsDecoded {
```
NumaryBot, line 38: 🔴 **[blocker] Update stale schemas even when structure is current**

When a specs file has the current row layout but a missing or outdated `$schema`, this early return reports no change and leaves the file untouched. Thus `numscript test --migrate` does not perform the advertised schema migration; include `specs.Schema != SchemaURL` in the change decision before returning the original bytes.

internal/specs_format/migrate_v0024.go
```diff
@@ -0,0 +1,285 @@
...
+	return changed
+}
+
+func decodeBalanceColors(rows interpreter.Balances) bool {
+	changed := false
+	for i := range rows {
+		if rows[i].Color != "" {
+			continue
+		}
+		asset, color, ok := splitLegacyColoredAsset(rows[i].Asset)
```
NumaryBot, line 245: 🔴 **[blocker] Avoid rewriting ambiguous underscored assets**

For any current-shape file containing an asset such as `USD_RED` with no explicit color, this unconditionally rewrites it as asset `USD` plus color `RED`. The old format also permitted literal underscored asset names, so the source representation cannot distinguish those cases; `--migrate` can therefore silently change test semantics and overwrite the file. Ambiguous assets should be left unchanged or reported for manual resolution.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/182#issuecomment-5425452721

Azorlogh (approved):

internal/specs_format/migrate.go
ascandone: we don't want to actually fix the $schema field, as it's just an hint for the editor

internal/specs_format/migrate_v0024.go
ascandone: it's ok as is

PR #190, opened 2026-09-14
feat: replace the funds execution engine with internal/funds

ascandone:
Replaces the interpreter's funds execution engine with a new one in `internal/funds`.

> **Stacked on #191** (differential testing), which is itself stacked on #189 (spec fixtures). Review just the top commit; GitHub retargets this as the stack merges.

## This is a rewrite, not a move

An earlier version of this description called it an "extraction". That undersold it — the engine is new code and every script execution goes through it:

| | before | after |
|---|---|---|
| implementation | `fundsQueue` (188 lines) + `InternalBalances` (121) | `funds.RunState` (618) |
| API | `Push` / `PullAll` / `PullColored` / `Clone`; map + `fetchBalance` / `Set` / `Merge` | `Pull` / `PullUncapped` / `Send` / `ForcePosting` / `Save` / `MarkPush` / `MarkEnd` / `GetPostings` / `Prewarm` |
| rollback | snapshot and restore (`Clone` / `DeepClone`) | mark and rewind (`MarkPush` / `MarkEnd`) |
| balance cache | one amount per key | `{amount, baseLoaded}` — write deltas tracked apart from the fetched base |

`interpreter.go` goes **−254 / +133**: it loses its execution logic rather than gaining an import. It no longer generates postings at all — that's `rs.GetPostings()` now.

**Why:** this is the substrate the bytecode VM needs. It has to execute the same funds semantics without depending on the tree-walking interpreter, and the alternative is two implementations of these rules drifting apart.

## Evidence

The rewrite is checked against an independent implementation — the vendored legacy ledger machine — by the harness #191 adds:

- **255,033 fuzz executions on this branch, zero divergence** between the new engine and the oracle.
- The same harness on #191 alone (i.e. the *old* engine) gives 253,668 executions, also zero. So the baseline is established before the swap rather than after, and CI runs the harness on every PR.

**Coverage limit, stated plainly:** the generator emits sends, sources with overdrafts, allotments, `max`, `remaining`, `balance()`, `meta()` and `set_tx_meta`. It does **not** emit colored assets, scopes, `oneof`, or account interpolation — so the `oneof` rollback path, which behavior change 2 below depends on, is covered by the spec suite in #189 and not by the fuzzer.

## Behavior changes

Three, each pinned by a fixture here. Verified these are the only three: applying the four held-back fixtures to #189's tree fails exactly these and nothing else.

**1. `neg-max-dest`** — a negative amount in a destination `max` clause now fails with the existing `NegativeAmountErr` instead of emitting a posting.

**2. `scaling-with-oneof`** — a `oneof` branch that falls short now discards its scaled-swap postings. This is a direct consequence of the rollback mechanism change above. The new case is built so the swap is not a no-op (`acc1` holds a second scale), unlike the existing cases.

**3. `midscript-balance-after-credit`** — a mid-script `balance()` read is no longer masked by an earlier write to the same account.

The old `InternalBalances.has()` could not distinguish *"fetched from the store"* from *"created by a write"*, because `fetchBalance` created a zero entry on miss. So crediting `@alice` and then reading `balance(@alice)` saw a cache hit, skipped its fetch, and reported only what the script sent — dropping the starting balance. `funds.balanceEntry` now tracks `baseLoaded` separately, so `Has()` reports false for an entry holding only a write delta and `Prewarm` folds the fetched base into that delta rather than overwriting it.

The fixture has `@alice` start at 20 and receive 100: previously `USD/2 100`, now `USD/2 120`. `@carol` is the control — never written to, correct before and after — isolating this to write-masking rather than `balance()` generally.

**Not breaking:** mid-script calls are gated behind `experimental-mid-script-function-call`. `balance()` inside a `vars` block is unaffected, since vars resolve before execution and no write can mask them.

## Note for ledger implementors

The balance read is now a lazy fetch where there was none. In `TestMidscriptBalance`, `GetBalancesCalls` goes from `nil` to a single `{alice, USD/2}` fetch.

## Note for reviewers

`internal/specs_format` exports four previously package-private helpers (`MergeBalances`, `EndBalances`, `GetMovements`, `CompareMovements`). No consumer in this PR — a later PR's tooling uses them. Flagging so it isn't read as dead code.

## Verification

`go build ./...`, `go test ./...`, `gofmt -l` clean on the root module; difftest and oracle modules green; fuzz burst as above.


internal/funds/funds.go
```diff
@@ -0,0 +1,618 @@
...
+		if next.account != s.sources[i].account || next.scope != s.sources[i].scope || next.color != s.sources[i].color {
+			return
+		}
+		s.sources[i].amount.Add(s.sources[i].amount, next.amount)
+		s.removeAt(i + 1)
+	}
+}
+
+// removeAt preserves the order of the remaining sources.
+func (s *RunState) removeAt(i int) {
```
NumaryBot, line 626: 🟠 **[major] Avoid copying the queue tail on every consumed source**

For a send funded by N sources, each removal copies the remaining slice, making draining the queue O(N²) rather than the previous linked queue's O(N). Coalescing repeated sources in `compactAt` has the same problem because it repeatedly calls `removeAt(i + 1)`. Large source blocks therefore incur a substantial execution-time regression. Use a head cursor for FIFO consumption and compact retained entries in a single pass.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/190#issuecomment-5671549178

internal/interpreter/resolve_dependencies.go
```diff
@@ -78,7 +79,26 @@ func ResolveDependencies(ctx context.Context, store Store, vars map[string]strin
...
+	// balance() yields a Monetary, which can't name an account), so getBalance
+	// just hits the recording store and returns zero — no funds engine involved.
+	getBalance := func(account AccountAddress, asset Asset) (*big.Int, InterpreterError) {
+		_, err := recording.GetBalances(ctx, BalanceQuery{
+			{Account: account.Name, Asset: string(asset), Color: "", Scope: account.Scope},
+		})
+		if err != nil {
+			return nil, QueryBalanceError{WrappedError: err}
+		}
+		return new(big.Int), nil
```
NumaryBot, line 102: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🔴 **[blocker] Preserve balance values during dependency resolution**

With account interpolation and `get_amount` enabled, a balance can determine an account name: `monetary $b = balance(@treasury, USD)` followed by `number $id = get_amount($b)` can feed a source such as `@user:$id`. Returning zero instead of the fetched balance makes dependency resolution report `user:0` while execution accesses, for example, `user:100`. This breaks the dependency coverage contract for valid scripts. Preserve the fetched balance when evaluating origins rather than assuming monetary values cannot influence account names.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 1 comment-only finding.

These comments do not affect the active verdict or commit status.

ascandone:
Verified both findings rather than applying them blind. One was right and is fixed; the other is real but smaller than stated.

**Blocker — balance values during dependency resolution: confirmed and fixed** in `a6b8ba1`.

This was a regression introduced by this PR, not pre-existing. Repro:

```
vars {
  monetary $b = balance(@treasury, USD)
  number $id = get_amount($b)
}
send [USD 1] (
  source = @user:$id allowing unbounded overdraft
  destination = @out
)
```

with `@treasury` at 100:

| | resolved write | actual posting |
|---|---|---|
| `main` | `user:100` | `user:100` |
| this PR, before the fix | `user:0` | `user:100` |

On `main` resolution and execution shared one `getBalance` that returned the fetched value; this PR injected a stub returning zero, on the stated assumption that a `balance()` yields a Monetary and so can't name an account. `get_amount` breaks that assumption. A ledger pre-locking from the resolved set would have locked the wrong account.

Fixed by returning the fetched amount, and added to `TestResolveDependenciesCoversRuntime` — which asserts resolution covers what execution touches, so it's the right home. Confirmed the new case fails without the fix and passes with it.

Worth noting: the differential harness in #191 cannot catch this class at all, because `internal/gen` emits no account interpolation. That limit is stated in #191's description, and this is a concrete instance of it.

**Queue drain — real, but ~16% not an order of magnitude.** Measured, same script shape, N sources into one send, 20 runs:

| sources | `main` | this PR |
|---|---|---|
| 100 | 333 µs | 280 µs (16% *faster*) |
| 500 | 1.57 ms | 1.64 ms (5% slower) |
| 2000 | 10.3 ms | 11.9 ms (16% slower) |

`removeAt` does copy the tail, so the direction of the finding is right. But the previous engine was not O(N) end-to-end either — it goes 31x slower for 20x the sources, against 43x here — so this is a constant-factor regression at large source counts, not a complexity class change. Not fixing it in this PR: it's an optimization, it would mean re-touching the core of the engine this PR is already asking you to review, and the funds-queue head-cursor work belongs with the perf PR. Filed as a follow-up rather than silently dropped.

internal/funds/funds.go
```diff
@@ -0,0 +1,618 @@
...
+
+// ForcePosting moves amount from src to dst bypassing the funding queue, for
+// movements the queue does not model (e.g. asset-scaling conversions). Unlike Send
+// it uses the explicit asset argument, which may differ from the current asset. A
+// non-positive amount is a no-op; no balance sufficiency check is performed.
+//
+// Safe inside a region, unlike Send: it touches no queue entry, so a rewinding
+// MarkEnd undoes it completely by reversing the posting.
+func (s *RunState) ForcePosting(src, srcScope, dst, dstScope, asset, color string, amount *big.Int) error {
+	if amount.Sign() <= 0 {
```
NumaryBot: 🟠 **[major] Reject negative scaling postings instead of silently dropping them**

With asset scaling enabled, `FindScalingSolution` can return negative conversions for negative starting balances. For example, sending `[EUR 1]` from an account holding `EUR/2 -100` and `EUR/1 20` through a scaling account produces conversions of -100 and 20, with a net credit of EUR 1. Dropping only the negative conversion now lets execution succeed while debiting EUR 2 worth of positive funds for that EUR 1 credit; previously the negative posting caused an invariant error. Reject negative amounts or adjust the scaling solver so its net amount matches the conversions actually posted.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/190#issuecomment-5671549178

ascandone:
**Negative scaling postings: confirmed, and it was the more serious of the two. Fixed in `ad18b80`.**

Verified against `main` rather than taking it as read:

```
send [EUR 1] (
  source = @acc1 with scaling through @swap
  destination = @dest
)
```

with `@acc1` holding `EUR/2 -100` and `EUR/1 20`:

| | outcome |
|---|---|
| `main` | fails — `The script produced a posting with invalid values: {acc1 swap -100 EUR/2}` |
| this PR, before the fix | **succeeds**, posting `acc1 -> swap 20 EUR/1`, `swap -> acc1 1 EUR`, `acc1 -> dest 1 EUR` |

So `@acc1` gave up `EUR/1 20` — 2 EUR — to move 1 EUR, and `@swap` kept the difference. An error became a silently wrong money movement, which is the wrong direction for this to fail in.

The cause is as described: the scaling solution is the pair (-100, 20), netting the 1 EUR requested, and `ForcePosting` treated any non-positive amount as a no-op, so only the positive leg posted. `checkPostingInvariants` still exists and would have caught the negative posting — but the posting no longer reached it, because `ForcePosting` discarded it first. That's what made this invisible: the guard looks intact when you read it.

`ForcePosting` now returns `ErrNegativePosting` for a negative amount (zero stays a no-op) and the interpreter maps it to the same `InternalError`, so behavior matches `main` exactly, error message included. Pinned at both levels: `TestForcePosting_NegativeIsRejected` in `internal/funds`, and `TestScalingNegativeConversionIsRejected` end to end.

I deliberately did **not** touch `FindScalingSolution`. Whether it should produce such a solution at all is a real question, but answering it changes scaling semantics, whereas this restores an invariant. Worth a follow-up.

Neither this nor the dependency-resolution regression was reachable by the differential harness — the generator emits neither asset scaling nor account interpolation. Both are now covered by targeted tests instead.

Re-verified after the fix: root suite green, difftest green, and a fresh 117,225-execution fuzz run against the oracle with no divergence.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/190#issuecomment-5671549178

ascandone:
**On the red `codecov/project` check** — it's a project-threshold artifact, not a gap in the new engine. Checked rather than waved off:

- `codecov/patch` **passes**: 81.29% of the diff hit, against a 75.36% target. The new code is better covered than the repo average.
- `codecov/project` reports 74.67%, −0.70%.
- Measuring `internal/funds` with `-coverpkg` (the default per-package profile misreports it — `Has` and `AccountBalances` show 0% because their callers live in `internal/interpreter`): **87.6%, with every function above 80% except `AccountsMetadata.PrettyPrint`**, a debug helper.
- No dead code in the new package: `AssetToScaledAsset`, the other apparent 0%, is used by `batch_balances_query.go`.
- Lowest-covered functions among the files this PR touches are `sendToKeptOrDest` (60%), `evaluateColor` (67%), `resolveDestination` (69%) — error and edge branches, all pre-existing shapes rather than new engine logic.

So there's no uncovered new money-handling path. Happy to add tests for those branches if you'd rather see the project number green, but I didn't want to pad coverage to move a threshold.

PR #191, opened 2026-09-15
test: add differential testing against the legacy ledger machine

ascandone:
Adds differential testing: run randomly generated numscript programs through both this repo's interpreter and an independent implementation, and compare.

> **Stacked on #189** (spec fixtures). #190, the funds-engine rewrite, is stacked on this. Opened ahead of #190 deliberately — see below.

## What

| path | |
|---|---|
| `internal/oracle/` | the legacy ledger `machine`, vendored as close to verbatim as practical. Its own Go module, so it can never be reached from the published library. |
| `internal/gen/` | random numscript program generator, wired to Go's native fuzzing |
| `builder/` | AST builders the generator needs |
| `internal/difftest/` | the harness. Its own module, so neither the root module nor the oracle depends on it. |

## Why ahead of #190

#190 replaces the interpreter's funds execution engine. Without this, that rewrite's only witness is the repo's own spec suite — tests written by the same people changing the engine. The oracle is a separate implementation maintained elsewhere, so agreement with it is evidence that isn't self-referential.

Landing it first also establishes the baseline *before* the swap rather than after: **253,668 fuzz executions on this branch, against the current interpreter, with no divergence.** On #190 the same harness gives 255,033 executions, also clean. Those two numbers mean something only in that order.

## CI

`internal/oracle` and `internal/difftest` are separate Go modules, so the root module's `go test ./...` never reaches them — I checked, `go list ./...` returns zero packages from either. Landing the harness without wiring would have put it in the repo where it never runs, which would defeat the point.

So this adds a `Difftest` job: both modules' tests, plus a 60s fuzz burst. **The burst is a deliberate trade** — the committed corpus is only 3 seeds, so the regression tests alone are weak, and the burst is where divergences actually get found. It also means a genuine divergence can surface as a red check on an unrelated PR. That seemed right for this repo, but it's the one decision here worth a second opinion.

## Scope

Oracle-vs-interpreter only. The upstream harness also compares the compiler+VM, but `internal/vm` isn't on `main` yet, so `run_vm.go` and the VM legs are dropped here. `Compare()` is engine-agnostic and takes labels, so re-adding them is mechanical; the spot is marked in `difftest.go`.

`DIFFTEST_HANDOFF.md` is not included — it's working notes that describe themselves as not for commit.

## Coverage limits

The generator emits sends, sources with overdrafts, allotments, `max`, `remaining`, `balance()`, `meta()` and `set_tx_meta`. It does **not** emit colored assets, scopes, `oneof`, or account interpolation. Postings are compared on (source, destination, asset, amount); scopes and colors are not.

`Compare` also tolerates some legitimate disagreement — notably, it compares missing-funds *classification* rather than exact error text, because a source-side negative `max` clause is a known, deliberately unfixed gap. `regression_test.go` pins both sides of that: the tolerated case, and the companion case that must still be caught.

## Note on the vendored oracle

~11.5k of the ~14.6k lines added are `internal/oracle`, including generated ANTLR. It's a fork and will rot; it's isolated as a separate module so it can't reach the published library. Provenance and licensing are worth a look as part of this review. `machine/vm/machine.go` is intentionally left un-gofmt'd to stay close to the original.

## Verification

All three modules build and test green. Fuzz baseline as above.


NumaryBot (commented):
NumaryBot posted 3 new inline findings.

Summary: https://github.com/formancehq/numscript/pull/191#issuecomment-5677076804

internal/gen/extra.go
```diff
@@ -0,0 +1,406 @@
...
+// picked at a deliberate elevated rate (~1 in 4) specifically to exercise
+// balance(@world, ASSET) — confirmed directly against the oracle to be a
+// legal, always-zero read (never an error).
+func genVarDecls(rng *rand.Rand, poolSize int, balances map[BalanceKey]*big.Int, metadata map[MetaKey]string) []VarDecl {
+	n := rng.Intn(4) // 0..3
+	if n == 0 {
+		return nil
+	}
+
+	fundedKeys := make([]BalanceKey, 0, len(balances))
```
NumaryBot, line 140: 🔴 **[blocker] Sort map-derived keys before selecting with the seeded RNG**

When generated balances or metadata contain multiple entries, Go's randomized map iteration changes the order of `fundedKeys` and `metaKeys`, so the same fuzz bytes can select different origins and produce different programs. This breaks corpus replay and failure minimization in `FuzzDiff`, potentially making a saved divergence disappear on rerun. Sort both key slices before indexing them with the seeded RNG.

builder/origin.go
```diff
@@ -0,0 +1,64 @@
...
+// declareOriginVar lazily registers a `<typ> $name = <origin>` declaration
+// the first time the returned expression is rendered, and reuses the same
+// name on subsequent renders (mirroring ExprVar's dedup-by-identity, but
+// here identity is just "this closure", since each call site gets its own).
+func declareOriginVar[T ExprType](typ string, origin render) Expression[T] {
+	allocated := false
+	var name string
+
+	return func(env *env, w int) {
+		if !allocated {
```
NumaryBot, line 22: 🟠 **[major] Scope origin-variable allocation to each build environment**

Reusing an origin expression or a statement containing one across two `BuildProgram` calls leaves `allocated` true after the first build. The second build then emits `$originvar_0` without registering its declaration in the new environment; with other origin expressions it can instead reference an unrelated declaration with the same name. Track expression identity in the environment, as ordinary variables do, rather than retaining allocation state in the closure.

builder/expression.go
```diff
@@ -47,6 +60,46 @@ func ExprNumberBigInt(amount *big.Int) Expression[ExprTypeNumber] {
...
+func ExprAdd(a, b Expression[ExprTypeNumber]) Expression[ExprTypeNumber] {
+	return func(env *env, w int) {
+		a(env, w)
+		env.builder.WriteString(" + ")
+		b(env, w)
+	}
+}
+
+// ExprSub renders `<a> - <b>`.
+func ExprSub(a, b Expression[ExprTypeNumber]) Expression[ExprTypeNumber] {
```
NumaryBot, line 80: 🟠 **[major] Preserve right-hand expression grouping in subtraction**

When `b` is another arithmetic expression, concatenating its rendering changes the requested computation: `ExprSub(10, ExprSub(3, 1))` renders `10 - 3 - 1`, which evaluates to 6 rather than 8. The generator already constructs nested right-hand expressions through `genNumExpr`. Preserve the expression semantics through sign-aware flattening or another grammar-supported representation; the same issue affects nested `ExprMonetarySub`.

.github/workflows/main.yml
```diff
@@ -44,3 +44,25 @@ jobs:
     secrets:
       NUMARY_GITHUB_TOKEN: ${{ secrets.NUMARY_GITHUB_TOKEN }}
       GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+
+  Difftest:
```
NumaryBot, line 48: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Set an explicit timeout for the differential-testing job**

If dependency downloads, compilation, or a test process stalls, this job can occupy a runner until GitHub's default 360-minute job timeout. The `-fuzztime 60s` flag only bounds the fuzzing phase, not the entire job. Add a job-level `timeout-minutes` appropriate for both module test suites and the fuzz burst.

builder/expression.go
```diff
@@ -47,6 +60,46 @@ func ExprNumberBigInt(amount *big.Int) Expression[ExprTypeNumber] {
...
+	return func(env *env, w int) {
+		a(env, w)
```
NumaryBot, line 82: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Preserve grouping when rendering subtraction operands**

A nested right operand changes meaning when flattened: `ExprSub(10, ExprSub(3, 1))` renders `10 - 3 - 1`, which HEAD's left-associative grammar evaluates to 6 instead of 8. The same defect affects `ExprMonetarySub`. Preserve operand grouping (HEAD supports parentheses), or otherwise render an equivalent expression rather than silently changing the arithmetic.

builder/builder.go
```diff
@@ -114,18 +115,69 @@ func renderVar[T comparable](
...
+// writeStringLiteral writes s as a raw double-quoted numscript string
+// literal directly into env.builder, bypassing the vars pool (mirrors
+// UnsafeAccount's rationale: some grammar positions, like a meta key or a
+// set_tx_meta/set_account_meta key, only ever take a literal, never a
+// variable).
+func writeStringLiteral(env *env, s string) {
+	env.builder.WriteByte('"')
+	for i := 0; i < len(s); i++ {
+		ch := s[i]
+		if ch == '"' || ch == '\\' {
```
NumaryBot, line 127: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Preserve metadata keys under the parser's string semantics**

For keys containing a backslash or quote, this escaping changes the actual metadata key. HEAD's `parseStringTk` only removes the surrounding quotes and does not unescape the contents, so a single backslash becomes two, and a quote gains a literal backslash. Consequently `NewNumberVarFromMeta` looks up a different key and the metadata statements write a different key than requested. Align literal encoding with parser decoding and verify round-trip key values rather than only rendered snapshots.

internal/difftest/regression_test.go
```diff
@@ -0,0 +1,224 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := context.Background()
+			newRes := runNew(ctx, tc.script, nil, tc.balances, nil)
+			oracleRes := runOracle(ctx, tc.script, nil, tc.balances, nil)
+
+			v := Compare(newRes, oracleRes, "new interpreter", "oracle")
```
NumaryBot, line 139: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Assert rejection explicitly in the negative-destination regression**

The `negative max-clause amount` case passes even though HEAD's `sendTo` still clamps a negative destination cap to zero. The interpreter succeeds and the oracle rejects, but `Compare` deliberately tolerates success versus non-missing-funds errors. Consequently this test does not catch the very defect it claims to pin. Add explicit expected runtime outcomes for this case rather than relying solely on the relaxed differential verdict.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 4 comment-only findings.

These comments do not affect the active verdict or commit status.

ascandone:
Checked each finding rather than applying them blind.

**Blocker — map-derived key pools: confirmed and fixed** in `a1fd7ab`.

Quantified it first: with the pools unsorted, **51 of 400 seeds generated a different program from run to run**. That is exactly as damaging as the finding says — the fuzzer saves a failure as raw seed bytes and the shrinker re-runs those bytes expecting the same program back, so a saved divergence could simply fail to reproduce.

Fixed by sorting both pools, and added `TestGenerationIsSeedDeterministic` so it can't regress. The test sweeps many seeds deliberately: the hazard only appears once a map holds two or more entries, and my first single-seed probe passed 200 times in a row before a sweep found it.

**CI job timeout: fixed** in the same commit. `timeout-minutes: 15` — correct that `-fuzztime` bounds only the fuzzing phase, not module downloads or builds.

**Deferred, with reasons:**

*`ExprSub` / `ExprMonetarySub` grouping* — the rendering is genuinely wrong (`ExprSub(10, ExprSub(3,1))` renders `10 - 3 - 1`). But both engines execute the same rendered text, so it cannot produce a false divergence; it narrows what the generator can express rather than making the comparison unsound. Worth fixing, not worth blocking the harness on.

*Origin-variable allocation scoping* — real if a builder is reused across `BuildProgram` calls. `GenerateScript` builds fresh each time, so it isn't reachable today. Filing rather than fixing, since the fix is a change to `builder`'s environment model.

*Metadata key escaping* — the round-trip mismatch is real for keys containing a backslash or quote. `metaKeyPool` is `k0`/`k1`/`k2`, so nothing generated reaches it.

*Negative-destination regression assertion* — correct that the case currently passes vacuously, but that's a sequencing artifact rather than a defect here: on this branch the interpreter still clamps a negative destination cap, and #190 is the PR that makes it error. Adding the explicit outcome assertion now would make this PR red. It belongs in #190, where the behavior actually changes, and `neg-max-dest` there already pins it.

NumaryBot (commented):

Resolved 3 stale NumaryBot review threads (3 fixed, 0 outdated).

Summary: https://github.com/formancehq/numscript/pull/191#issuecomment-5677076804

internal/oracle/machine/script/compiler/source.go
```diff
@@ -0,0 +1,265 @@
...
+			if err != nil {
+				return nil, LogicError(c, err)
+			}
+		}
+	case *parser.SrcAllotmentContext:
+		if isAll {
+			return nil, LogicError(c, errors.New("cannot take all balance of an allotment source"))
+		}
+		p.PushAddress(*monAddr)
+		p.VisitAllotment(c.SourceAllotment(), c.SourceAllotment().GetPortions())
```
NumaryBot, line 39: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Propagate source-allotment compilation errors**

For a source allotment whose portions do not total 100% (for example, two 40% portions), `VisitAllotment` returns an error before emitting `OP_MAKE_ALLOTMENT`. Ignoring that error still emits `OP_ALLOC`, so compilation succeeds with malformed bytecode and execution panics when it expects an allotment on the stack. Return the compilation error here so the differential harness receives a compile failure rather than crashing.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 1 comment-only finding.

These comments do not affect the active verdict or commit status.

NumaryBot (commented):

Resolved 1 stale NumaryBot review thread (1 fixed, 0 outdated).

Summary: https://github.com/formancehq/numscript/pull/191#issuecomment-5677076804

internal/difftest/run_new.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, p := range execResult.Postings {
+		postings = append(postings, Posting{
+			Source:      p.Source,
+			Destination: p.Destination,
+			Asset:       p.Asset,
+			Amount:      p.Amount,
+		})
+	}
+
+	return SideResult{Postings: postings}
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Compare generated metadata outputs**

Both adapters retain only postings, discarding transaction and account metadata. The generator emits `set_tx_meta` and `set_account_meta`, including arithmetic and `meta()`-derived values, but incorrect keys, values, or dropped writes therefore pass comparison whenever postings agree. Preserve normalized metadata in `SideResult` and compare it for successful executions so these generated statements actually detect behavioral divergences.

internal/gen/extra.go
```diff
@@ -0,0 +1,481 @@
...
+// balances (genPresetBalances) are the only way to make a balance()-origin
+// var observe a non-zero value.
+func genSeedStatements(rng *rand.Rand, poolSize int, assets []string) Program {
+	stmts := make(Program, 0, poolSize*len(assets))
+	for i := range poolSize {
+		for _, asset := range assets {
+			stmts = append(stmts, Statement{
+				IsSendAll: false,
+				Amount:    Monetary{Asset: asset, AssetAsVar: asVar(rng), Amount: seedAmount(rng)},
+				Source:    Source{Kind: SrcAccount, Account: "world", AccountAsVar: asVar(rng)},
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Keep seed-funding sources as literal @world**

Randomizing `AccountAsVar` here makes any affected funding statement cause the oracle’s `ResolveBalances` to reject the entire script before execution. `runOracle` classifies that as a compile failure, which `Compare` tolerates without comparing postings. Each funded script contains 6–45 funding statements, each independently choosing this form with probability 1/3, so most funded cases bypass the intended differential check. Emit literal `@world` for setup funding and retain world-variable exploration in the dedicated account-variable cases.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 2 comment-only findings.

These comments do not affect the active verdict or commit status.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/191#issuecomment-5677076804

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/191#issuecomment-5677076804

internal/difftest/run_new.go
```diff
@@ -0,0 +1,124 @@
...
+			Key:     k.Key,
+			Value:   value,
+		})
+	}
+
+	// Defensive copy: vars is shared with runOracle's call in RunOne, and
+	// nothing here should depend on whether this function mutates its
+	// input (it doesn't today, but that's not a documented guarantee).
+	execResult, err := parseResult.Run(ctx, maps.Clone(vars), store)
+	if err != nil {
```
NumaryBot, line 93: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Preserve interpreter invariant failures as internal errors**

The interpreter already returns `interpreter.InternalError` when `checkPostingInvariants` detects an invalid posting. This adapter converts it into an ordinary `RunErr` with `MissingFunds=false`. If the oracle succeeds, `Compare` therefore tolerates the failure and reports agreement, hiding precisely the invalid-posting regressions this harness should catch. Classify interpreter internal errors as `SideResult.InternalErr` before handling ordinary runtime rejections, and test that conversion.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 1 comment-only finding.

These comments do not affect the active verdict or commit status.

builder/builder.go
```diff
@@ -114,18 +128,71 @@ func renderVar[T comparable](
...
 	env *env,
 ) string {
 
+	// Render every origin var's RHS to a string FIRST: doing so may lazily
+	// register new entries into env's account/string/asset/number pools
+	// (e.g. an asset literal not otherwise referenced in the program body),
+	// which must be visible to the renderVar calls below — otherwise a
+	// pool entry discovered only here would never get its own `type $name`
+	// declaration line.
+	originRHS := make([]string, len(env.originVars))
```
NumaryBot: 🟠 **[major] Resolve nested origin declarations before sizing their render buffer**

Passing a `NewAccountVarFromMeta` expression as the account of `NewMonetaryVarFromBalance` causes rendering the outer origin to append another entry to `env.originVars`. This range loop does not visit that new entry, and the subsequent declaration loop indexes beyond `originRHS`, panicking during `BuildProgram`. Discover origin dependencies transitively and render them in dependency order before emitting declarations.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/191#issuecomment-5677076804

internal/difftest/compare.go
```diff
@@ -0,0 +1,284 @@
...
+		// (destination, asset) must still agree exactly.
+		//
+		// This is a real loss of checking power, not a normalization: in a
+		// script using `kept`, a genuinely wrong SOURCE would be tolerated
+		// here. It is scoped as narrowly as the script text allows, and
+		// Verdict.Tolerated makes each occurrence countable.
+		if usesKept(script) {
+			aByDest := aggregateByDestination(aRes.Postings)
+			bByDest := aggregateByDestination(bRes.Postings)
+			if !destTotalsDiffer(aByDest, bByDest) {
```
NumaryBot: 🟠 **[major] Compare metadata before returning a kept-attribution tolerance**

When a successful `kept` script has different source attribution but matching destination totals, this early return skips both metadata comparisons below. A wrong or dropped `set_tx_meta` or `set_account_meta` write therefore passes whenever that tolerated posting difference also occurs. Retain the tolerance result but return it only after metadata has been checked.

NumaryBot (commented):

Summary: https://github.com/formancehq/numscript/pull/191#issuecomment-5677076804

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 3 comment-only findings.

These comments do not affect the active verdict or commit status.

.github/workflows/main.yml
```diff
@@ -44,3 +44,26 @@ jobs:
...
+      # differential harness would sit in the repo and never run.
+      - name: oracle module
+        working-directory: internal/oracle
+        run: go test ./...
+      # Includes TestDifferentialSweep, the deterministic differential gate: a
+      # fixed set of generated programs run through both engines. The
+      # coverage-guided FuzzDiff is deliberately NOT run here — its -fuzztime
+      # deadline surfaces as a test failure with no divergence behind it, so it
+      # belongs in a scheduled job or a local session, not a PR gate.
+      - name: difftest regressions and differential sweep
```
NumaryBot, line 67: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Establish a passing baseline before enabling the differential gate**

This command includes `TestDifferentialSweep`, which HEAD explicitly marks as expected to fail on existing semantic disagreements. The sweep treats every mismatch as fatal, while `TestKnownOpenDivergences` confirms those disagreements remain. Consequently, the new job fails on the baseline instead of identifying regressions. Resolve the baseline disagreements or explicitly account for known cases while preserving failures for new divergences before enabling this gate.

builder/builder.go
```diff
@@ -114,18 +130,120 @@ func renderVar[T comparable](
...
+	visit = func(id int) {
+		if seen[id] {
+			return
+		}
+		// Marked before recursing: an origin cannot reference itself
+		// transitively (each is rendered once, and a reference can only be
+		// to a var allocated during that render), but a cycle here would
+		// otherwise be a stack overflow rather than a merely wrong order.
+		seen[id] = true
+		for _, m := range originVarRefRe.FindAllStringSubmatch(originRHS[id], -1) {
```
NumaryBot, line 190: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Ignore quoted metadata keys when ordering origin declarations**

The regex treats `$originvar_N` inside a metadata key as a dependency. For example, render `NewAccountVarFromMeta(World(), "$originvar_1")` first, then a balance origin using that account. The resulting RHS expressions make this traversal emit declaration 1 before declaration 0, although balance actually depends on declaration 0. The interpreter then fails with an unbound-variable error. Record dependencies during expression rendering, or exclude string literals when scanning references, and add a regression covering this valid metadata key.

internal/difftest/sweep_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	var sb strings.Builder
+	fmt.Fprintf(&sb, "%d divergence class(es) over %d seeds:\n", len(classes), sweepSeeds)
+	for _, k := range order {
+		cl := classes[k]
+		fmt.Fprintf(&sb, "  %-44s %4d scripts (first: seed %d)\n", k, cl.count, cl.firstSeed)
+	}
+	sb.WriteString("\nSee internal/oracle/DIVERGENCES.md §3.\n")
+
+	first := classes[order[0]]
```
NumaryBot, line 90: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Establish a passing baseline for the required differential sweep**

The sweep unconditionally fails on known disagreements that HEAD deliberately preserves, including save and subsequent-use-of-kept behavior pinned by `TestKnownOpenDivergences`. The new `Difftest` CI job runs this test through `go test ./...`, so its documented “expected to fail” state leaves the required gate red before any regression occurs. Resolve those disagreements or explicitly baseline the known cases while continuing to fail on new differences.

NumaryBot (commented):

Resolved 1 stale NumaryBot review thread (1 fixed, 0 outdated).

Summary: https://github.com/formancehq/numscript/pull/191#issuecomment-5677076804

ascandone:
Superseded by a three-PR split, same final tree (verified byte-identical via `git diff`):

- #193 `builder/`
- #194 `internal/gen/` — the seed-driven generator
- #195 `internal/oracle/` + `internal/difftest/` + the Difftest CI job

The branch `test/difftest-oracle-harness` is kept, not deleted, because #190 is still based on it. #190 should be retargeted onto `test/difftest-oracle` (#195).

PR #195, opened 2026-09-17
test: differential testing against the legacy ledger machine

ascandone:
## What changed

Runs generated programs through both the numscript interpreter and a vendored
copy of the ledger repo's `internal/machine`, and compares what moved.

- **`internal/oracle/`** — that vendored copy, in its own module so it cannot
  drift into the interpreter's dependency graph. `smoke_test.go` checks the
  vendoring itself did not change behaviour. `DIVERGENCES.md` records every
  difference from ledger main and why it is there; `DIFFERENCES-BY-EXAMPLE.md`
  is the same list as runnable scripts.
- **`internal/difftest/`** — the harness, also its own module: `Compare` (what
  counts as a real difference vs a posting-granularity one), the seeded
  `TestDifferentialSweep` gate, `TestKnownBugRepros` / `TestKnownOpenDivergences`,
  and `FuzzDiff` with its corpus.
- **CI gets a `Difftest` job.** Both are separate modules, so the root
  `go test ./...` reaches neither and without this the harness would sit in the
  repo and never run. It runs the deterministic sweep, not the fuzzer:
  `-fuzztime` expiry surfaces as a test failure with no divergence behind it,
  which belongs in a scheduled job, not a PR gate.

## Why

Third of three PRs replacing #191. Stacked on #194 (generator) → #193 (builder).
It exists so the funds-engine swap (#190) has an independent baseline rather
than being reviewed on assertion alone.

**Base is `feat/program-generator`** — retargets as the stack merges.

## Risk

**MEDIUM**, and not for the usual reason: nothing here ships. The risk is that
the harness looks more authoritative than it is. Two things bound it, both
written down rather than discovered later:

1. **The generator does not reach asset scaling, account interpolation or
   colors.** A green sweep says nothing about them.
2. **The oracle is not fully independent.** Two deliberate departures from
   ledger main, in `DIVERGENCES.md` §2 — the `OP_TAKE` negative-amount guard
   (① upstream agrees it is a bug; PR #2060 closed unmerged) and `kept`
   consuming its funding (②).

## Validation

- [x] Baseline validation: `go build ./...`, `go vet ./...`, gofmt clean in all
      three modules
- [x] Targeted tests: `go test ./...` in `internal/oracle` and `internal/difftest`
- [x] Full suite / broader validation: `TestDifferentialSweep` green —
      **0 divergence classes and 0 tolerated over 3000 seeds**

## Architecture / behavior impact

Two new Go modules, neither reachable from the root module's build. One new CI
job. No interpreter, API or storage change.

## Review focus

**`DIVERGENCES.md` §2 ② is where judgment is most valuable.** The sweep went
green partly by changing the oracle: ledger does not consume a `kept` funding —
it returns to the pool and funds the next destination — while the interpreter
does. The oracle now consumes it, matching the interpreter, at two sites in
`script/compiler/destination.go`.

That removed 7 failing and 85 tolerated scripts and let `Compare`'s
`kept source attribution` tolerance be deleted outright — a real gain, since it
would have swallowed a genuinely wrong *source* in any script mentioning `kept`
and short-circuited before metadata was compared.

The cost is that **the oracle can no longer catch a `kept` regression in the
interpreter.** `TestOracleKeptComplex` keeps ledger's original expectations in
its comment as the compensating control. Worth disagreeing with if you think
the oracle should have stayed faithful and the sweep stayed red.

Second: `Compare`'s remaining relaxations. Each one is checking power traded for
signal, and they are the difference between a harness that finds bugs and one
that agrees with itself.

## Known concerns

`TestKnownOpenDivergences` pins two numscript↔ledger differences that are real
and undecided — `save` past the balance (§3.2) and source-side negative `max`
(§3.3, §3.4). They assert the divergence is *still there*, so they fail loudly
if anyone changes the semantics without updating the doc. Neither is an oracle
defect and neither blocks this PR.


internal/difftest/sweep_test.go
```diff
@@ -0,0 +1,116 @@
...
+// and the generator churns it hard (hundreds of new interesting inputs a
+// minute). A fixed seed set has no deadline, no corpus, and no workers, so a
+// failure here always means a real disagreement — and names the seed that
+// reproduces it.
+//
+// Run FuzzDiff directly for exploration:
+//
+//	go test -run '^$' -fuzz FuzzDiff -fuzztime 5m ./internal/difftest
+//
+// EXPECTED TO FAIL right now. The oracle is faithful to ledger everywhere
```
NumaryBot: **🟡 [minor] Stale "EXPECTED TO FAIL" sweep comment**

Comment claims the sweep should fail with s3 divergences visible, but DIVERGENCES.md and the PR report it green (0 classes, 0 tolerated). Update the comment to expect green.

internal/difftest/regression_test.go
```diff
@@ -0,0 +1,283 @@
...
+	if !newRes.MissingFunds {
+		t.Fatalf("expected the interpreter to fail specifically due to missing funds; got new=%+v", newRes)
+	}
+
+	if v := Compare(script, newRes, oracleRes, "new interpreter", "oracle"); !v.Mismatch {
+		t.Fatalf("expected new-vs-oracle to be flagged as a mismatch, got none")
+	}
+}
+
+// TestKnownOpenDivergences pins the numscript/ledger disagreements that are
```
NumaryBot, line 276: **🟡 [minor] Open-divergence pins misattributed**

Only s3.2 is pinned here; s3.3 is pinned in TestSourceSideNegativeMaxClauseTolerated and s3.4 is resolved, so the "faithful on all three" comment misroutes fixes. Cite the actual pinning tests.

internal/difftest/run_oracle.go
```diff
@@ -0,0 +1,121 @@
...
+		return entry
+	}
+	for k, amount := range balances {
+		getOrCreateEntry(k.Account).Balances[k.Asset] = new(big.Int).Set(amount)
+	}
+	for k, value := range metadata {
+		getOrCreateEntry(k.Account).Metadata[k.Key] = value
+	}
+	if err := m.ResolveResources(ctx, store); err != nil {
+		return SideResult{CompileErr: err.Error()}
```
NumaryBot, line 53: **🟡 [minor] Oracle resolution errors swallowed**

SetVarsFromJSON/ResolveResources/ResolveBalances errors become CompileErr, which Compare tolerates b-side, so a real oracle resolution divergence passes as ok() and metadata is never compared. Classify these separately.

internal/difftest/run_oracle.go
```diff
@@ -0,0 +1,121 @@
...
+	case machine.AccountAddress:
+		return string(x)
+	case machine.Asset:
+		return string(x)
+	case *machine.MonetaryInt:
+		return (*big.Int)(x).String()
+	case machine.Monetary:
+		return string(x.Asset) + " " + (*big.Int)(x.Amount).String()
+	default:
+		return fmt.Sprintf("%v", v)
```
NumaryBot, line 113: **🟡 [minor] %v fallback for unknown meta values**

The default branch %v-formats unlisted machine.Value types (e.g. Portion) despite the comment forbidding default formatting, risking false divergences. Cover every case or return an error.

internal/oracle/smoke_test.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+		20% to @qux
+		5% kept
+		remaining to @quz
+	}
+)`, store)
+	require.NoError(t, err)
+
+	require.Len(t, m.Postings, 6)
+	requirePosting(t, m.Postings[0], "foo", "arst", "GEM", 2)
```
NumaryBot, line 262: **🟡 [minor] kept divergence not machine-checked**

TestOracleKeptComplex, the stated compensating control for s2(ii), asserts the oracle's new expectations; ledger's original postings live only in its comment. Keep them in a skipped or golden assertion.

internal/difftest/sweep_test.go
```diff
@@ -0,0 +1,116 @@
...
+		if !seen {
+			cl = &class{firstSeed: seed, sampleWhy: c.OracleVsNew.Reason, sampleVars: c.Vars, sampleSrc: c.Script}
+			classes[k] = cl
+			order = append(order, k)
+		}
+		cl.count++
+	}
+
+	for what, n := range tolerated {
+		t.Logf("tolerated: %-28s %d/%d scripts", what, n, sweepSeeds)
```
NumaryBot: **🟡 [minor] Tolerated counts never bounded**

The sweep only t.Logf's tolerance counts though Tolerated's doc calls absorption a blind spot, so the gate stays green if a tolerance absorbs scripts. Fail above a fraction of sweepSeeds.

internal/difftest/compare.go
```diff
@@ -0,0 +1,217 @@
...
+	// number of posting lines (e.g. one engine emits a single merged
+	// posting where the other emits two that sum to the same amount,
+	// observed when an unbounded-overdraft account sits between two
+	// non-adjacent draws from the same real account in a source list).
+	// That's a posting-granularity difference, not a change in what moved
+	// where — so it's intentionally not treated as a mismatch here.
+	aAgg := aggregatePostings(aRes.Postings)
+	bAgg := aggregatePostings(bRes.Postings)
+
+	if postingsDiffer(aAgg, bAgg) {
```
NumaryBot, line 113: **🔵 [suggestion] Posting diff walk duplicated**

Compare calls postingsDiffer then repeats the same length/key/amount walk inline to pick a message, traversing both maps twice and needing the two to stay in sync. Return the differing key or inline it.

internal/difftest/sweep_test.go
```diff
@@ -0,0 +1,116 @@
...
+		first.firstSeed, first.sampleWhy, first.sampleVars, first.sampleSrc)
+	t.Fatal(sb.String())
+}
+
+// divergenceClass buckets a Verdict reason by its kind, discarding the
+// per-script amounts and account names that follow it, so the sweep can count
+// classes rather than listing thousands of individually-worded failures.
+func divergenceClass(reason string) string {
+	head, _, _ := strings.Cut(reason, "\n")
+	for _, prefix := range []string{
```
NumaryBot, line 158: **🔵 [suggestion] divergenceClass couples to message text**

divergenceClass hardcodes prefixes of the mismatch messages formatted in compare.go; rewording one degrades bucketing and inflates class counts. Share the class strings.

internal/difftest/compare.go
```diff
@@ -0,0 +1,217 @@
...
+	Tolerated string
+}
+
+func ok() Verdict { return Verdict{} }
+
+func tolerated(what string) Verdict { return Verdict{Tolerated: what} }
+
+func mismatch(format string, args ...any) Verdict {
+	return Verdict{Mismatch: true, Reason: fmt.Sprintf(format, args...)}
+}
```
NumaryBot, line 25: **🔵 [suggestion] Dead tolerance plumbing in Verdict**

`tolerated()` has no callers and `Verdict.Tolerated` can never be non-empty now that the `kept source attribution` tolerance was deleted, so the sweep's tolerance accounting can never observe a value. Keep the helper/field only if a tolerance is planned, otherwise remove them to avoid implying live checking-power tracking.

internal/oracle/machine/script/compiler/source.go
```diff
@@ -0,0 +1,265 @@
...
+				return nil, LogicError(c, err)
+			}
+		}
+	case *parser.SrcAllotmentContext:
+		if isAll {
+			return nil, LogicError(c, errors.New("cannot take all balance of an allotment source"))
+		}
+		p.PushAddress(*monAddr)
+		p.VisitAllotment(c.SourceAllotment(), c.SourceAllotment().GetPortions())
+		p.AppendInstruction(program.OP_ALLOC)
```
NumaryBot, line 42: **🟡 [minor] Swallowed CompileError from VisitAllotment in VisitValueAwareSource**

In `VisitValueAwareSource` (SrcAllotmentContext branch) the return value of `p.VisitAllotment(...)` is ignored before `OP_ALLOC` is emitted, so an invalid allotment (e.g. portions summing over 100%) would silently compile a malformed stack instead of failing. Today this helper has no caller in the oracle compiler package, so impact is latent, but the parallel path in `VisitMonetary` (compiler.go) correctly propagates the error. Either delete the dead helper or check and return the `*CompileError` for consistency.

NumaryBot (commented):
Findings

internal/oracle/machine/script/generate.sh
```diff
@@ -0,0 +1,19 @@
...
+if [[ "${TRACE-0}" == "1" ]]; then
+    set -o xtrace
+fi
+
+cd "$(dirname "$0")"
+
+ANTLR_VERSION='4.10.1'
+
+main() {
+  curl --continue-at - https://www.antlr.org/download/antlr-$ANTLR_VERSION-complete.jar -O
```
NumaryBot: **🔵 [suggestion] ANTLR jar download is resumed without integrity verification**

`curl --continue-at - -O` resumes any pre-existing (possibly truncated or corrupted) jar and the script never verifies the artifact, so a corrupt jar could silently feed `org.antlr.v4.Tool` during `go generate`. Consider removing stale downloads, or pinning and verifying a SHA-256 checksum for `antlr-4.10.1-complete.jar` before running the tool.

internal/difftest/run_oracle.go
```diff
@@ -0,0 +1,121 @@
...
+			store[account] = entry
+		}
+		return entry
+	}
+	for k, amount := range balances {
+		getOrCreateEntry(k.Account).Balances[k.Asset] = new(big.Int).Set(amount)
+	}
+	for k, value := range metadata {
+		getOrCreateEntry(k.Account).Metadata[k.Key] = value
+	}
```
NumaryBot, line 51: **🟡 [minor] Oracle resolution failures classified as compile errors**

`ResolveResources`/`ResolveBalances` failures (including "invalid program (resolve balances: …)") return `SideResult{CompileErr}`, so `Compare`'s tolerated b-side-rejection branch silently absorbs them, even though the harness rule says internal errors must mismatch unconditionally (spec; standards). Map resolution-stage failures — at minimum the "invalid program" class — to `RunErr` or `InternalErr` so they participate in the missing-funds comparison instead of the rejection tolerance.

internal/difftest/compare.go
```diff
@@ -0,0 +1,217 @@
...
+	// empty when the two sides agreed outright. It exists so a sweep can
+	// count how often each tolerance fires: a tolerance that starts
+	// absorbing most of the corpus has stopped being a tolerance and
+	// become a blind spot.
+	Tolerated string
+}
+
+func ok() Verdict { return Verdict{} }
+
+func tolerated(what string) Verdict { return Verdict{Tolerated: what} }
```
NumaryBot, line 21: **🟡 [minor] Dead tolerance plumbing: Verdict.Tolerated is never set**

Every tolerance path in `Compare` returns `ok()`, so `tolerated()` has no callers, `Verdict.Tolerated` is always empty, and `TestDifferentialSweep`'s tolerated counter can never fire — making "0 tolerated over 3000 seeds" trivially true rather than a measured signal, and contradicting the field's own doc (spec and standards agreed). Either return `tolerated(...)` from the b-side-rejection branch or delete the helper, field and sweep bookkeeping.

internal/difftest/sweep_test.go
NumaryBot: **🟡 [minor] Stale `EXPECTED TO FAIL` comment on the PR gate test**

The comment on `TestDifferentialSweep` claims the sweep is EXPECTED TO FAIL and reports §3 divergences, but `DIVERGENCES.md` and the PR both state the sweep is green (0 divergence classes over 3000 seeds), so the comment misleads maintainers about whether the `Difftest` CI job can pass (spec). Update it to say the sweep is expected green and that open divergences are pinned in `TestKnownOpenDivergences`/`TestSourceSideNegativeMaxClauseTolerated`.

internal/difftest/regression_test.go
```diff
@@ -0,0 +1,283 @@
...
+// real and still undecided — see internal/oracle/DIVERGENCES.md §3. None of
+// them is an oracle defect: the oracle is faithful to ledger on all three, and
+// each was previously invisible because the oracle had been bent toward
+// numscript to hide it.
+//
+// These assert that a mismatch IS still reported. If one starts passing,
+// something changed the semantics — update DIVERGENCES.md and move the case
+// into TestKnownBugRepros rather than deleting it.
+func TestKnownOpenDivergences(t *testing.T) {
```
NumaryBot, line 283: **🟡 [minor] TestKnownOpenDivergences pins only the save divergence**

`TestKnownOpenDivergences` contains only the §3.2 `save` case while its comment (and the PR description) claim it pins the §3.3/§3.4 negative-`max` divergences too; the source-side case lives in `TestSourceSideNegativeMaxClauseTolerated`/`TestMissingFundsClassificationMismatchStillCaught`, and §3.4 is resolved and pinned nowhere (spec). Align the comment (and PR description) with the code or add the missing pins.

internal/difftest/compare.go
```diff
@@ -0,0 +1,217 @@
...
+	aAgg := aggregatePostings(aRes.Postings)
+	bAgg := aggregatePostings(bRes.Postings)
+
+	if postingsDiffer(aAgg, bAgg) {
+		if len(aAgg) != len(bAgg) {
+			return mismatch(
+				"aggregated posting set differs: %s has %d distinct (source,destination,asset), %s has %d\n%s: %+v\n%s: %+v",
+				aLabel, len(aAgg), bLabel, len(bAgg), aLabel, aRes.Postings, bLabel, bRes.Postings,
+			)
+		}
```
NumaryBot, line 119: **🔵 [suggestion] Duplicated map walk in posting comparison**

`postingsDiffer(aAgg, bAgg)` traverses both maps, then the `len(aAgg) == len(bAgg)` branch immediately re-walks `aAgg` against `bAgg` to build the mismatch message (standards). Fold both into a single pass that returns the differing key to avoid duplicated logic that can drift.

internal/difftest/cmd/rundiff/main.go
```diff
@@ -0,0 +1,62 @@
...
+
+	for i := range *n {
+		rngSeed := *seed + int64(i)
+		rng := rand.New(rand.NewSource(rngSeed))
+
+		c := difftest.RunOne(ctx, rng)
+
+		if *verbose {
+			fmt.Printf("--- seed %d ---\n%s\n\n", rngSeed, c.Script)
+		}
```
NumaryBot, line 40: **🔵 [suggestion] Single-element loop scaffolding for a second engine that doesn't exist**

`rundiff`, `FuzzDiff` and `regression_test.go` iterate one-element slices (`[]difftest.Verdict{c.OracleVsNew}`, single-pair slices) in anticipation of a second engine leg (standards). Inline the single element until a real second engine exists, then reintroduce the loop.

internal/difftest/compare.go
```diff
@@ -0,0 +1,217 @@
...
+	// MissingFunds classification, not raw fail/succeed, is the thing that
+	// must agree — see SideResult.MissingFunds's doc comment. A known,
+	// deliberately-unfixed gap (a negative source-side `max ... from`
+	// clause amount is a hard reject on the oracle, but silently
+	// contributes zero on the interpreter, which can let a *different*
+	// source in the same list cover the shortfall instead) means fail-vs-
+	// succeed alone is too strict: it flags that gap's known, accepted
+	// divergence in what's *accepted* as a false positive. Whether the
+	// funds were actually adequate is the invariant this harness cares
+	// about, so that's what's compared instead.
```
NumaryBot: **🟡 [minor] Non-funds fail-vs-succeed divergence is tolerated in both directions**

When one side fails at runtime with a non-missing-funds error and the other succeeds, `Compare` returns `ok()` because the `MissingFunds` classifications match (both false), so any genuine divergence — e.g. the interpreter erroring on a script the oracle executes, with postings and metadata never compared — passes silently. The tolerance is only needed for the known source-side negative-`max` gap (oracle rejects, interpreter accepts). Consider narrowing it to that known pattern (or to b-side-rejects direction only) so an unexpected fail-vs-succeed divergence on either side is still reported.

internal/oracle/machine/script/compiler/source.go
```diff
@@ -0,0 +1,265 @@
...
+			if err != nil {
+				return nil, LogicError(c, err)
+			}
+		}
+	case *parser.SrcAllotmentContext:
+		if isAll {
+			return nil, LogicError(c, errors.New("cannot take all balance of an allotment source"))
+		}
+		p.PushAddress(*monAddr)
+		p.VisitAllotment(c.SourceAllotment(), c.SourceAllotment().GetPortions())
```
NumaryBot: **🟡 [minor] Ignored `VisitAllotment` error in `VisitValueAwareSource`**

In the `SrcAllotmentContext` arm of `VisitValueAwareSource`, the `*CompileError` returned by `p.VisitAllotment(...)` is discarded, so an invalid allotment (e.g. portions summing over 100%) silently emits `OP_ALLOC` and reports compile success. Sibling call sites such as `VisitMonetary` in `compiler.go` check this error and abort. The method currently has no callers in package `compiler`, making this a latent defect — handle the error via `LogicError` or delete the dead method.

internal/oracle/machine/script/generate.sh
NumaryBot: **🟡 [minor] ANTLR jar downloaded without integrity check or curl --fail**

`curl --continue-at - -O` saves any HTTP error body as `antlr-4.10.1-complete.jar` and never verifies a checksum, so a failed or truncated download yields a corrupt/tampered jar that `java` then runs for codegen. Add `--fail` (and ideally `--proto '=https'` with a pinned SHA-256 check) before invoking the tool.

internal/oracle/machine/value.go
```diff
@@ -0,0 +1,86 @@
...
+	TypeNumber                     // 64bit unsigned integer
+	TypeString                     // string
+	TypeMonetary                   // [asset number]
+	TypePortion                    // rational number between 0 and 1 both inclusive
+	TypeAllotment                  // list of portions
+	TypeAmount                     // either ALL or a SPECIFIC number
+	TypeFunding                    // (asset, []{amount, account})
+)
+
+func (t Type) String() string {
```
NumaryBot, line 22: **🟡 [minor] Type.String() omits TypeFunding and TypeAmount**

The `switch` in `Type.String()` covers `TypeAccount`..`TypeAllotment` but has no case for `TypeFunding` or `TypeAmount`, so both print as "invalid type" — e.g. `OP_DELETE` in `vm/machine.go` formats a `TypeFunding` value with `%v` in its "wrong type" error. Add the missing cases so diagnostics name the actual type.

NumaryBot (commented):
Findings

NumaryBot (commented):
Findings

internal/difftest/cmd/rundiff/main.go
```diff
@@ -0,0 +1,62 @@
...
+
+		for _, v := range []struct {
```
NumaryBot, line 42: **🔵 [suggestion] single-element verdict loop is unused scaffolding**

The mismatch loop iterates a one-entry slice of {pair, verdict} structs, but no second engine or Case.VM leg exists yet, so the indirection only adds noise around the single real check on c.OracleVsNew. Compare the field directly, or keep the loop only once a second engine leg actually lands.

internal/difftest/run_new.go
```diff
@@ -0,0 +1,116 @@
...
+			Account: k.Account,
+			Key:     k.Key,
+			Value:   value,
+		})
+	}
+
+	// Defensive copy: vars is shared with runOracle's call in RunOne.
+	execResult, err := parseResult.Run(ctx, maps.Clone(vars), store)
+	if err != nil {
+		_, missingFunds := err.(numscript.MissingFundsErr)
```
NumaryBot, line 91: **🟡 [minor] missing-funds classification uses a bare type assertion**

`err.(numscript.MissingFundsErr)` matches only an unwrapped concrete type, so a wrapped missing-funds error would silently classify as false. Compare's run-failure tolerance keys on this classification, so degradation would surface as spurious `missing-funds classification differs` sweep failures. Use `errors.As` instead.

internal/difftest/regression_test.go
```diff
@@ -0,0 +1,269 @@
...
+    @acc3 allowing unbounded overdraft
+  }
+  destination = @acc1
+)`
+
+	ctx := context.Background()
+	newRes := runNew(ctx, script, nil, nil, nil)
+	oracleRes := runOracle(ctx, script, nil, nil, nil)
+
+	for _, pair := range []struct {
```
NumaryBot, line 228: **🟡 [minor] vestigial single-element loop in TestSourceSideNegativeMaxClauseTolerated**

One `Compare` call is wrapped in a one-entry anonymous `pair` slice loop; the sibling entries were deleted along with the old `kept` tolerance. Call `Compare` directly so the assertion reads plainly.

internal/difftest/run_oracle.go
```diff
@@ -0,0 +1,115 @@
...
```
NumaryBot, line 51: **🟡 [minor] resolve-stage failures folded into `CompileErr`**

`SetVarsFromJSON`, `ResolveResources` and `ResolveBalances` errors are returned as `CompileErr`, so an oracle resolution failure (missing meta key, negative preset balance) takes `Compare`'s tolerated b-side-rejection path even when the interpreter executed the script fine, silently skipping the comparison. Comment why resolution failures share the compile tolerance, or return a distinct stage.

internal/difftest/compare.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return ok()
+}
+
+// compareMeta compares one normalized metadata map as a set of key/value pairs.
+func compareMeta(what string, a, b map[string]string, aLabel, bLabel string) Verdict {
+	for k, av := range a {
+		bv, present := b[k]
+		if !present {
+			return mismatch("%s: %s wrote %q=%q, %s wrote nothing for that key\n%s: %v\n%s: %v",
```
NumaryBot, line 150: **🔵 [suggestion] `compareMeta` duplicates its missing-key branch**

The two loops repeat the same "wrote nothing for that key" mismatch construction with swapped labels. Extract one helper taking writer/other maps and labels and call it twice.

internal/gen/ast.go
```diff
@@ -179,27 +161,23 @@ const (
...
-// (OriginAccountBalanceContext/OriginAccountMetaContext).
+// VarDecl is a `vars {}` declaration whose value comes from the compiler, not
+// from a runtime binding.
 type VarDecl struct {
 	Kind    VarDeclKind
 	Account string
-	// Render Account through a var rather than inline, so that
-	// `balance($accountN, ...)` and a literal `@acc` elsewhere in the script
-	// become two resources aliasing one account.
+	// Makes `balance(, ...)` and a literal `@acc` elsewhere alias one
```
NumaryBot: **🟡 [minor] Garbled doc comment on VarDecl.AccountAsVar**

The comment now reads "Makes `balance(, ...)` and a literal `@​acc` elsewhere alias one account" — the diff dropped the var placeholder that previously read `balance($accountN, ...)`, leaving a malformed example. Replace with `balance($var, ...)` so the aliasing example stays understandable.

internal/difftest/compare.go
```diff
@@ -0,0 +1,215 @@
...
+		// a mismatch -- see TestMissingFundsClassificationMismatchStillCaught.
+		if aNegativeAmount(aRes) && bRes.MissingFunds {
+			return tolerated("negative amount vs missing funds")
+		}
+		return mismatch(
+			"missing-funds classification differs: %s missingFunds=%v (runErr=%q), %s missingFunds=%v (runErr=%q)",
+			aLabel, aRes.MissingFunds, aRes.RunErr, bLabel, bRes.MissingFunds, bRes.RunErr,
+		)
+	}
+	if aRunFailed != bRunFailed {
```
NumaryBot, line 93: **🟠 [major] Any one-sided runtime failure is tolerated, not just the known negative-max divergence**

`if aRunFailed != bRunFailed { return ok() }` accepts every case where one engine errors and the other commits with matching `MissingFunds` classification, not only DIVERGENCES.md #4's negative-`max` rejection. A real bug where one engine fails while the other moves money never surfaces in the sweep. Narrow it to the known shape, e.g. require the failing side's `RunErr` to be the negative-amount rejection.

internal/oracle/DIVERGENCES.md
```diff
@@ -0,0 +1,262 @@
...
+
+```sh
+L=~/Documents/dev/formance/ledger/internal/machine
+O=internal/oracle/machine
+norm() { sed -E 's|github\.com/formancehq/ledger/internal/machine|M|g; s|github\.com/formancehq/numscript/internal/oracle/machine|M|g' "$1"; }
+for f in $(cd $O && find . -name '*.go' | sed 's|^\./||' | sort); do
+  [ -f "$L/$f" ] && diff <(norm "$L/$f") <(norm "$O/$f") > /dev/null || echo "DIFFERS: $f"
+done
+```
+
```
NumaryBot, line 362: **🟡 [minor] Drift table says only destination.go changes behaviour, contradicting test comments**

`internal/difftest/regression_test.go` credits fixes for `StaticStore.GetBalances` and `UnresolvedResourceBalances` to `internal/oracle/machine/vm/store.go` and `vm/machine.go` (both confirmed present in the snapshot), while the drift table classifies those files as vendoring-only and claims "Only one is a behaviour change". If the fixes are absent from ledger main at `8880965c2`, the file's own record-drift rule is violated. Reconcile the table or cite the upstream source.

internal/difftest/compare.go
```diff
@@ -0,0 +1,215 @@
...
+func mismatch(format string, args ...any) Verdict {
+	return Verdict{Mismatch: true, Reason: fmt.Sprintf(format, args...)}
+}
+
+// Compare normalizes and diffs two engines' results for the same script.
+// aLabel/bLabel appear in mismatch messages only.
+//
+// Error strings are never compared, only whether an error occurred and at which
+// stage: wording legitimately differs between implementations.
+func Compare(script string, aRes, bRes SideResult, aLabel, bLabel string) Verdict {
```
NumaryBot, line 32: **🟡 [minor] `Compare` ignores its `script` parameter**

The `script` parameter is never referenced in the body or in mismatch messages, and all callers pass the real script. Drop it, or include it in mismatch output to aid debugging.

NumaryBot (commented):
Findings

internal/difftest/compare.go
```diff
@@ -0,0 +1,215 @@
...
+	for k, av := range a {
+		bv, ok := b[k]
+		if !ok || av.Cmp(bv) != 0 {
+			return true
+		}
+	}
+	return false
+}
+
+// aNegativeAmount reports whether this side rejected the script for a negative
```
NumaryBot, line 211: **🟡 [minor] Negative-amount classification relies on error-text substring matching**

`aNegativeAmount` tests `strings.Contains(ToLower(RunErr), "negative amount")`. A reworded interpreter error silently converts a tolerated known divergence into sweep noise, and an unrelated a-side error mentioning the phrase would be tolerated as funds-vs-negative. Match the typed error instead (e.g. `errors.As` on numscript's negative-amount error type).

internal/difftest/compare.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Aggregated by (source, destination, asset) rather than compared as an
+	// ordered list: the engines can split the same net transfer into a different
+	// number of posting lines, seen when an unbounded-overdraft account sits
+	// between two non-adjacent draws from the same account. That is granularity,
+	// not a change in what moved where.
+	aAgg := aggregatePostings(aRes.Postings)
+	bAgg := aggregatePostings(bRes.Postings)
+
+	if postingsDiffer(aAgg, bAgg) {
```
NumaryBot, line 113: **🔵 [suggestion] Posting difference computed twice**

`postingsDiffer(aAgg, bAgg)` walks both maps, then the mismatch branch re-walks them to build the message; the two passes can drift. Extract one helper returning the first differing key and use it for both the check and the message.

internal/difftest/regression_test.go
```diff
@@ -0,0 +1,269 @@
+package difftest
```
NumaryBot, line 1: **🟡 [minor] TestKnownOpenDivergences pins only one of the two claimed open divergences**

The PR description says `TestKnownOpenDivergences` pins both undecided divergences (save past balance and source-side negative max), but the test contains only the save case; the negative-max case is pinned by `TestSourceSideNegativeMaxClauseTolerated` and `TestMissingFundsClassificationMismatchStillCaught`, which DIVERGENCES.md's pinning table already states correctly. Impact is description drift, not missing coverage. Update the description to name the tests that actually pin each case.

.github/workflows/main.yml
```diff
@@ -44,3 +44,26 @@ jobs:
     secrets:
       NUMARY_GITHUB_TOKEN: ${{ secrets.NUMARY_GITHUB_TOKEN }}
       GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+
+  Difftest:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
```
NumaryBot, line 53: **🟡 [minor] Submodule dependencies are never cached in the Difftest job**

actions/setup-go@​v5 caches only the root go.sum by default; internal/oracle and internal/difftest are separate modules with their own go.sum, so their dependencies are re-downloaded on every CI run. Add cache-dependency-path listing both module go.sum files.

internal/difftest/compare.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+		return mismatch(
+			"missing-funds classification differs: %s missingFunds=%v (runErr=%q), %s missingFunds=%v (runErr=%q)",
+			aLabel, aRes.MissingFunds, aRes.RunErr, bLabel, bRes.MissingFunds, bRes.RunErr,
+		)
+	}
+	if aRunFailed != bRunFailed {
+		// Neither side's (possible) failure was a missing-funds one — see
+		// above. Tolerated.
+		return ok()
```
NumaryBot: **🟡 [minor] Tolerated outcomes returned via ok() escape Tolerated blind-spot accounting**

Verdict.Tolerated exists so the sweep can count how often each tolerance fires, but the b-side compile rejection, run-failure asymmetry, and both-failed paths return ok() with an empty Tolerated, so TestDifferentialSweep's tolerance log never sees them. Return named tolerances so over-absorption stays measurable.

internal/oracle/machine/script/compiler/source.go
NumaryBot: **🟡 [minor] `VisitAllotment` error silently ignored in `VisitValueAwareSource`**

`p.VisitAllotment(c.SourceAllotment(), ...)` returns a `*CompileError` (e.g. portions summing over 100%) that is discarded, unlike the parallel `SrcAllotment` path in `VisitMonetary` which propagates it. An invalid allotment would compile to garbage instead of reporting an error. Either check and return the error, or delete the helper if it is unreachable dead code (no call sites exist in the `compiler` package).

internal/oracle/machine/script/generate.sh
NumaryBot: **🟡 [minor] `curl` lacks `--fail` when fetching the ANTLR jar**

`curl --continue-at - ... -O` exits 0 on HTTP errors (404/5xx) and writes the error-page body into `antlr-4.10.1-complete.jar`, leaving a corrupt jar so the subsequent `java -cp` run fails with a confusing class-not-found. Add `--fail` (and ideally `-L` and a checksum pin) so download failures abort `go generate`.

internal/oracle/DIVERGENCES.md
```diff
@@ -0,0 +1,373 @@
...
+| 1 | negative send amount | `insufficient funds` | `insufficient funds` | rejects the amount |
+| 2 | `kept` | does not consume the funding | consumes it | consumes it |
+| 3 | `save` past the balance | balance goes negative | floors at zero | floors at zero |
+| 4 | source-side negative `max` | rejects the script | rejects the script | clause contributes zero |
+| 5 | negative bounded overdraft cap | applies it as-is | applies it as-is | clamps it to zero |
+
+The oracle matches ledger on 1, 4 and 5. It does not on 2 and 3, the two places
+it cannot be trusted to check numscript.
+
+`TestDifferentialSweep` is green: 0 divergences over 3000 seeds, 106 of those
```
NumaryBot, line 24: **🟡 [minor] Tolerated-script counts for divergence #1 contradict each other**

The header says "106 of those scripts tolerated under #1" while §1 says "currently 161 of 3000 scripts, 5.4%" for the same tolerance, and the PR description's validation claims 0 tolerated over 3000 seeds. At most one of these describes the sweep at this SHA. Re-run the sweep and align the doc and description on one number.

internal/difftest/regression_test.go
```diff
@@ -0,0 +1,322 @@
...
+	if !newRes.MissingFunds {
+		t.Fatalf("expected the interpreter to fail specifically due to missing funds; got new=%+v", newRes)
+	}
+
+	if v := Compare(script, newRes, oracleRes, "new interpreter", "oracle"); !v.Mismatch {
+		t.Fatalf("expected new-vs-oracle to be flagged as a mismatch, got none")
+	}
+}
+
+// TestKnownOpenDivergences pins the numscript/ledger disagreements that are
```
NumaryBot, line 276: **🟡 [minor] TestKnownOpenDivergences does not pin the divergences the description claims**

The PR description's known-concerns section says this test pins save-past-the-balance and source-side negative max, asserting each divergence is still present, but it contains only the negative bounded overdraft cap case (DIVERGENCES.md #5). A silent change to save or negative-max semantics would not fail this test. Add the two cases or correct the description.

internal/gen/scenario.go
```diff
@@ -0,0 +1,340 @@
...
+	setupSave, setupSaveAll, setupFund, setupDrainPlain, setupDrawPlain,
+	setupDrawBounded, setupDrainBounded, setupDrawUnbounded, setupKeptSplit,
+}
+
+var observerKinds = []string{
+	observeDrainPlain, observeDrainBounded, observeDrawPlain, observeDrawBounded,
+	observeDrawFallback, observeDrawCapped,
+}
+
+// shadow is the generator's running estimate of the focus balance. It only
```
NumaryBot, line 46: **🔵 [suggestion] shadow.exact is written but never read**

`scenarioGen.sh.exact` is initialized in `genScenario` and cleared by `setupKeptSplit`, `observeDrawFallback` and `observeDrawCapped`, but no code reads it, so the documented "estimate did not see it" behavior is not implemented. Either use it to gate `boundaryAmount` anchoring or drop the field.

internal/difftest/compare.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	return false
+}
+
+// aNegativeAmount reports whether this side rejected the script for a negative
+// amount rather than for the funds.
+func aNegativeAmount(a SideResult) bool {
+	return !a.MissingFunds && a.RunErr != "" &&
+		strings.Contains(strings.ToLower(a.RunErr), "negative amount")
+}
```
NumaryBot, line 215: **🟡 [minor] Negative-amount classification by error-text substring**

`aNegativeAmount` gates the `negative amount vs missing funds` tolerance via `strings.Contains` on `RunErr`, though `Compare` documents that error strings are never compared; a wording change in either engine silently flips that script class between tolerated and mismatch. Classify with a typed flag on `SideResult` (e.g. `errors.As` with `machine.ErrNegativeAmount`), like `MissingFunds`.

internal/difftest/compare.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		total.Add(total, p.Amount)
+	}
+	return agg
+}
+
+func postingsDiffer(a, b map[postingKey]*big.Int) bool {
+	if len(a) != len(b) {
+		return true
+	}
```
NumaryBot, line 201: **🔵 [suggestion] `postingsDiffer` duplicates `Compare`'s diff walk**

`postingsDiffer` repeats the length check and per-key `Cmp` that the mismatch branch of `Compare` immediately redoes to build its message. Duplicated Code: have it return the first differing key so both paths share one walk.

internal/difftest/cmd/rundiff/main.go
```diff
@@ -0,0 +1,62 @@
...
+	for i := range *n {
+		rngSeed := *seed + int64(i)
+		rng := rand.New(rand.NewSource(rngSeed))
+
+		c := difftest.RunOne(ctx, rng)
+
+		if *verbose {
+			fmt.Printf("--- seed %d ---\n%s\n\n", rngSeed, c.Script)
+		}
+
```
NumaryBot, line 41: **🔵 [suggestion] Speculative single-engine loop scaffolding**

The one-element `[]struct{pair, verdict}` loop (mirrored by `Case.AnyMismatch` and a one-entry pair loop in `regression_test.go` line 228) exists only for a second engine that does not exist yet — Speculative Generality. Inline until the VM leg actually lands.

internal/difftest/sweep_test.go
```diff
@@ -0,0 +1,173 @@
...
+	row("...in that order", r.inOrder)
+	row("...where the save also overdraws", r.overdraws)
+	row("...that diverged", r.inOrderDiverg)
+	return sb.String()
+}
+
+// divergenceClass buckets a Verdict reason by its kind, discarding the
+// per-script amounts and account names that follow it, so the sweep can count
+// classes rather than listing thousands of individually-worded failures.
+func divergenceClass(reason string) string {
```
NumaryBot, line 156: **🔵 [suggestion] `divergenceClass` couples sweep grouping to message wording**

`divergenceClass` prefix-parses `Verdict.Reason`, so rewording a `Compare` mismatch message silently merges or splits sweep classes. Add a stable machine-readable kind to `Verdict` and group on that instead.

internal/difftest/run_oracle.go
```diff
@@ -0,0 +1,115 @@
...
+		return string(x)
+	case machine.Asset:
+		return string(x)
+	case *machine.MonetaryInt:
+		return (*big.Int)(x).String()
+	case machine.Monetary:
+		return string(x.Asset) + " " + (*big.Int)(x.Amount).String()
+	default:
+		return fmt.Sprintf("%v", v)
+	}
```
NumaryBot, line 114: **🟡 [minor] `normalizeOracleMetaValue` default branch guesses formatting**

Unhandled `machine.Value` types fall through to `fmt.Sprintf("%v", v)`, so a value type the generator can produce (e.g. a `Portion`) would be compared against the interpreter's rendering using oracle-specific formatting, risking a spurious metadata mismatch. Surface the unhandled type as an internal error instead of guessing.

internal/difftest/compare.go
```diff
@@ -0,0 +1,217 @@
...
+			// filter these upstream, but that makes this comparison
+			// silently dependent on an invariant it doesn't enforce —
+			// enforce it here too so the equivalence lives in one place.
+			continue
+		}
+		k := postingKey{Source: p.Source, Destination: p.Destination, Asset: p.Asset}
+		total, ok := agg[k]
+		if !ok {
+			total = new(big.Int)
+			agg[k] = total
```
NumaryBot, line 191: **🟡 [minor] Posting aggregation cancels opposite-direction movements**

`aggregatePostings` sums postings per `(source,destination,asset)`, so a side posting `A→B 10` plus `B→A 4` aggregates to the same net 6 as a single `A→B 6`, hiding a real gross-movement divergence between the engines. Consider tracking signed direction (e.g. aggregate source-minus-destination per ordered pair, or fail on opposite-direction keys) instead of pure netting.

internal/gen/ast.go
```diff
@@ -179,27 +161,23 @@ const (
 	MetaPortion
 )
 
-// VarDecl is a `vars {}` declaration whose value comes from the compiler,
-// not a runtime-supplied binding. See builder.NewMonetaryVarFromBalance /
-// builder.NewNumberVarFromMeta and internal/oracle's VisitVars
-// (OriginAccountBalanceContext/OriginAccountMetaContext).
+// VarDecl is a `vars {}` declaration whose value comes from the compiler, not
```
NumaryBot, line 164: **🔵 [suggestion] Comment lost the var placeholder in `balance(, ...)`**

The rewritten `AccountAsVar` comment now reads "Makes `balance(, ...)` and a literal `@​acc` ...", having dropped the `$accountN` placeholder. Restore it as `balance($accountN, ...)` so the aliasing example stays intelligible.

internal/oracle/machine/script/compiler/source.go
NumaryBot: **🟡 [minor] VisitValueAwareSource ignores allotment compile errors**

The *CompileError returned by p.VisitAllotment is discarded, so an invalid allotment (e.g. portions summing over 100%) would silently emit OP_ALLOC, unlike VisitMonetary in script/compiler/compiler.go which propagates the error. The method has no call sites in the compiler package, so the impact is currently latent; return or handle the error.

NumaryBot (commented):
Findings

internal/oracle/DIVERGENCES.md
```diff
@@ -0,0 +1,373 @@
...
+through to the insufficient-funds branch.
+
+Upstream agrees it is a bug. `vm/machine_negative_amount_test.go` pins the
+current behaviour with a `BUG:` comment. The fix, ledger PR #2060, was closed
+unmerged and its branch deleted, so main still has no guard.
+
+**The oracle copies ledger exactly, missing guard and all.** `Compare` absorbs the
+difference instead, under the name `negative amount vs missing funds`. Both
+engines reject the script and no money moves either way; only the error differs.
+
```
NumaryBot, line 61: **🟡 [minor] Contradictory tolerated-count for the #1 tolerance**

The summary (line 24) and 'What the sweep compares' say the `negative amount vs missing funds` tolerance fired 106 of 3000 scripts, while section #1 says 161 of 3000, 5.4%, and the PR description claims 0 tolerated. One count is stale, undermining the tolerance's stated blind-spot tracking. Re-run `TestDifferentialSweep`, align doc and description on one number. (spec + standards)

NumaryBot (commented):
Findings

internal/difftest/regression_test.go
```diff
@@ -0,0 +1,322 @@
...
+// real and still undecided — internal/oracle/DIVERGENCES.md. Neither is an
+// oracle defect: the oracle is faithful to ledger on both.
+//
+// These assert that a mismatch IS still reported. If one starts passing,
+// something changed the semantics — update DIVERGENCES.md and move the case
+// into TestKnownBugRepros rather than deleting it.
+func TestKnownOpenDivergences(t *testing.T) {
+	testCases := []struct {
+		name     string
```
NumaryBot, line 285: **🟡 [minor] PR description misstates what TestKnownOpenDivergences pins**

At this head its `testCases` holds only the negative bounded overdraft cap (DIVERGENCES.md #5); `save` past the balance is pinned by `TestKnownBugRepros` and the negative-`max` cases by dedicated tests, per DIVERGENCES.md's pin table. Update the description or restore the claimed cases. (spec)

internal/difftest/run_new.go
```diff
@@ -0,0 +1,122 @@
...
```
NumaryBot, line 91: **🟡 [minor] Error classification uses direct type assertions, not errors.As**

`_, missingFunds := err.(numscript.MissingFundsErr)` and the `NegativeAmountErr` assertion fail silently if the interpreter ever wraps those errors, making `Compare` misclassify a real divergence through the #1/#4 tolerances. Use `errors.As` for both. (standards)

internal/difftest/compare.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	// Aggregated by (source, destination, asset) rather than compared as an
+	// ordered list: the engines can split the same net transfer into a different
+	// number of posting lines, seen when an unbounded-overdraft account sits
+	// between two non-adjacent draws from the same account. That is granularity,
+	// not a change in what moved where.
+	aAgg := aggregatePostings(aRes.Postings)
+	bAgg := aggregatePostings(bRes.Postings)
+
```
NumaryBot, line 112: **🔵 [suggestion] Duplicated posting-comparison logic**

`postingsDiffer(aAgg, bAgg)` re-implements the same len-plus-key-loop that `Compare` then repeats to build the mismatch message, and `compareMeta` repeats the same a/b walk shape. Extract one comparison returning the differing key and reuse it. (standards)

.github/workflows/main.yml
```diff
@@ -44,3 +44,26 @@ jobs:
     secrets:
       NUMARY_GITHUB_TOKEN: ${{ secrets.NUMARY_GITHUB_TOKEN }}
       GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+
```
NumaryBot, line 47: **🔵 [suggestion] New modules get tests in CI but no vet/gofmt gate**

The `Difftest` job runs only `go test ./...` in the two new modules, and the root go-pr/go-test workflows cannot reach them, so vet/gofmt on `internal/oracle` and `internal/difftest` is enforced nowhere despite DIVERGENCES.md citing a gofmt exclusion for `/oracle/`. Add a vet/gofmt step to the job. (standards)

.github/workflows/main.yml
NumaryBot, line 53: **🔵 [suggestion] setup-go cache does not cover the new submodule go.sum files**

The Difftest job runs `go test` in `internal/oracle` and `internal/difftest`, whose dependencies live in those modules' own `go.sum` files, but `actions/setup-go@​v5` defaults to caching only the root `go.sum`. Pass `cache-dependency-path: internal/*/go.sum` so module downloads are cached instead of re-fetched every run.

PR #196, opened 2026-09-18
refactor(gen): trim comments and unexport the internal-only helpers

ascandone:
## What changed

Comment cleanup in `internal/gen`, plus three unexports. No behaviour change —
`-238/+146`, all of it comments and three identifier renames.

- **Haskell provenance removed.** The package was a port, and the comments still
  narrated it: `Gen.hs`/`Numscript.hs`/`Utils.hs`, "mirrors QuickCheck's
  `frequency`", "mirrors Haskell's `Ratio Int`", "a port of Utils.hs's
  cleanupSrcs". None of that is a fact about this code.
- **Comments that restate their field** (`// Render Asset through a var rather
  than inline.` on `AssetAsVar`).
- **Rationale compressed** where it had grown past being read — the 16-line
  `AccountVarDecl` doc, the 10-line `monetary` doc.
- **`GenerateScriptAST`, `ToBuilder`, `ToBuilderScript` unexported.** Nothing
  outside the package used them; `GenerateScript` and `RandFromBytes` are the
  entry points.

Kept: what cannot be read off the code — why `max`-clause amounts are never
negative, why both pools are sorted before being indexed with the seeded rng,
why the var-collision shapes are deliberately biased for.

## Why

Follow-up to #194, which merged before this cleanup was ready. Two of the
changes are fixes rather than tidying:

- Several comments referenced `DIFFTEST_HANDOFF.md`, which is not in the repo.
  Pointed at `internal/oracle/DIVERGENCES.md` instead, which is.
- `genPresetMetadata` carried two doc comments, the first superseded by the
  second.

## Risk

**LOW** — comments and three unexports of identifiers with no external callers.

## Validation

- [x] Baseline validation: `go build ./...`, `go vet ./...`, gofmt clean
- [x] Targeted tests: `go test -count=1 ./internal/gen/`
- [x] Full suite / broader validation: root `go test -count=1 ./...` green, and
      `internal/difftest`'s `TestDifferentialSweep` still green against it
      (3000 seeds, 0 divergences) — that is the real check that generation is
      unchanged, since the sweep is a pure function of the generator's output

## Architecture / behavior impact

N/A. The unexported names had no callers outside `internal/gen`; the external
test package (`gen_test`) uses the AST types and `GenerateProgram`, which stay
exported.

## Review focus

Whether anything cut was load-bearing. The judgment call throughout was
"provenance and restatement go, non-derivable facts stay" — the places worth
checking are `monetary`'s never-negative rule and `genVarDecls`' sort-before-index
note, both of which I kept but shortened.

## Known concerns

None.


ascandone:
Folded into #195 instead — see https://github.com/formancehq/numscript/pull/195/commits/a24638c. This only existed because #194 merged mid-cleanup; #195 is the better home.

